PR #5, opened 2024-10-04
feat: provide style code backbone

LouisJamesVasseur:
Define a color scheme, text size and style conventions and a light/dark mode to create the basic style of the app

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/theme/Theme.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/theme/Theme.kt
@@ -1,0 +1,46 @@
+package com.github.se.cyrcle.ui.theme
+
+import android.os.Build
+import androidx.compose.foundation.isSystemInDarkTheme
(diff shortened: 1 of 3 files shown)


sean-prz:
Be mindful to respect the name convention for branches as defined by @andrepeiry, thank you

app/src/main/java/com/github/se/cyrcle/ui/theme/Theme.kt
andrepeiry, line 1: If I wanted to be picky, the line 19-27 would be `//` comment, with only the actual comment being between `/* */` (it could also be a `//`, but the first type would further differentiate the title vs content).

andrepeiry (approved):
Simple but good code. It is standard to have a blank line at the end of each file, but this was not specified in our guidelines, just a sidenote.

Pijukai:
Good to go for me.

Pijukai (approved):

PR #7, opened 2024-10-05
cicd: implement a basic CI.yaml file to test

andrepeiry:
Implement the CI with GitHub Actions.
This PR should be squashed into a single commit, as this might take some trial and errors (so a lot of "test" commits)
The GitHub Action doesn't succeed because of a file not respecting the KTfmt.


Pijukai (approved):
Just need to format `MainActivity.kt` for further testing but seems like it runs well.

PR #10, opened 2024-10-07
Feature navigation

Armetiis:
This PR introduces a navigation system for our Android app using Jetpack Compose Navigation. The implementation is based on the navigation structure we developed during the bootcamp.

### Key Changes
Added NavigationActions.kt with route definitions and navigation logic
Updated MainActivity.kt with NavHost setup and navigation graph structure. Used placeholders for future screen setups

### Testing
Unit tests with mocking implemented for NavigationActions class

### Next Steps
Implement content for each screen (Auth, List, Map, Card)
Integrate navigation with app functionality


LouisJamesVasseur (approved):
This all seems good to me! It's coherent w/ the bootcamp structure so it should work well here too. It will have to be gradually updated as we add more screens

PR #12, opened 2024-10-07
MapScreen and MapBox implementation and API KEY Management

Pijukai:
This is the pull request for the MapBox Api implementation and the implementation of a simple map in `MapScreen.kt` . 
A few dependencies were added, to check that they are correct i invite you to follow the links below which indicates the procedure : 
https://docs.mapbox.com/android/maps/guides/install/

I also invite you to instanciate `MapScreen()` inside `setContent{}` of the `MainActivity.kt` to test properly the map. 

This PR also updates the `gitignore` file which aim to protect the public MapBox API KEY. On this point, i also need your advice about how to manage this API KEY since it does not follow the usual format. Hence i did not committed the Public API key which is needed for the Map to work. 

Note that the MapBox API implemented here support just the Map, navigation and other features are not part of this API but can be build upon it by adding their dependencies.


andrepeiry:
> it is not very useful at the moment.
Is it required at all at some moment ? Is there a way to ask it only when really necessary (temporary permission) ?

Pijukai:
It is not required for now but will be for the next Sprint since we will probably address the user localisation permission during it. So would be better to let it pass sonar cloud.

app/build.gradle.kts
andrepeiry, line 1: Suspicious space at line 150 between `implementation` and the following `(`
`implementation (...` should be replace with `implementation(...`

app/src/main/res/values/mapbox_access_token.xml
andrepeiry, line 1: This file can be removed, as it is a API key that should not be on the repository.

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -0,0 +1,25 @@
...
+      Modifier.fillMaxSize(),
+      mapViewportState =
+          rememberMapViewportState {
+            setCameraOptions {
+              zoom(2.0)
+              center(Point.fromLngLat(-98.0, 39.5))
+              pitch(0.0)
+              bearing(0.0)
+            }
+          },
```
andrepeiry, line 23: Suspicious and lonely `,` at line 23.

andrepeiry (changes requested):
Don't forget that a blank line at the end of a file is always nicer than none, thank you for the `MapScreen` example.

app/src/main/res/values/mapbox_access_token.xml
Pijukai: Done !

Pijukai:
@andrepeiry I added the blank spaces at the end of the necessary files but it seems that github doesn't show correctly.

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Pijukai: Fixed

andrepeiry (approved):
Everything is to my taste, thank for the quick fix

andrepeiry:
> @andrepeiry I added the blank spaces at the end of the necessary files but it seems that github doesn't show them correctly.

GitHub won't show you the blank line, but it will show a sign if it is missing

Pijukai:
@andrepeiry  Just add the secret and i think the CI will work. The last commit secured the API KEYs for MapBox in `local.properties`.

LouisJamesVasseur:
We may need to give further thought/discussion to whether or not we need to give fine-grain location access for further features of our app. For now I agree that it is not needed and coarse-grain is sufficient. Style-based comments were provided by @andrepeiry and I haven't seen any others worth noting. I tested the map and the functioning seems correct

LouisJamesVasseur (approved):
Everything seems in order and the map displays as expected, so I'll approve this pull request

PR #13, opened 2024-10-08
fix: correct NavHost declaration

matttsss:
Replace hardcoded string with Screen.* values in NavigationActions.kt
Fix brackets in NavHost structure declaration:
    First `navigation` block encompassed all following `navigation` blocks

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/navigation/NavigationActions.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/navigation/NavigationActions.kt
@@ -31,8 +31,10 @@ data class TopLevelDestination(val route: String, val icon: ImageVector, val tex
 
 /** Object containing the top level destinations in the app. */
 object TopLevelDestinations {
-  val LIST = TopLevelDestination(route = Route.LIST, icon = Icons.Outlined.Menu, textId = "List")
-  val MAP = TopLevelDestination(route = Route.MAP, icon = Icons.Outlined.LocationOn, textId = "Map")
+  val LIST =
+      TopLevelDestination(route = Route.LIST, icon = Icons.Outlined.Menu, textId = Screen.LIST)
+  val MAP =
+      TopLevelDestination(route = Route.MAP, icon = Icons.Outlined.LocationOn, textId = Screen.MAP)
 }
 
 /** List of top level destinations in the app. */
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/se/cyrcle/ui/navigation/NavigationActions.kt
```diff
@@ -31,8 +31,10 @@ data class TopLevelDestination(val route: String, val icon: ImageVector, val tex
 
 /** Object containing the top level destinations in the app. */
 object TopLevelDestinations {
-  val LIST = TopLevelDestination(route = Route.LIST, icon = Icons.Outlined.Menu, textId = "List")
-  val MAP = TopLevelDestination(route = Route.MAP, icon = Icons.Outlined.LocationOn, textId = "Map")
+  val LIST =
+      TopLevelDestination(route = Route.LIST, icon = Icons.Outlined.Menu, textId = Screen.LIST)
```
Armetiis, line 35: I believe that if one wanted to use a predefined String for the textId, it would be better to opt for Route.LIST ("List") rateher than Screen.LIST ("List Screen"), in the context that this is the text that will be displayed on the buttons (or menu) that allow going from one TLD to another

app/src/main/java/com/github/se/cyrcle/ui/navigation/NavigationActions.kt
```diff
@@ -31,8 +31,10 @@ data class TopLevelDestination(val route: String, val icon: ImageVector, val tex
...
+  val MAP =
+      TopLevelDestination(route = Route.MAP, icon = Icons.Outlined.LocationOn, textId = Screen.MAP)
```
Armetiis, line 37: The above argument is also valid here

Pijukai:
CI is good and so am I !

Armetiis (commented):
Thank you for noticing and fixing the indentation problem in the NavHost declaration. 
Read my comment below regarding the changes in the Top Level Destinations declaration.

Apart from that, all good

Pijukai (approved):
We can merge as is. Do we wait for the other Screens in order to push a complete `MainActivity.kt`  or will it be the topic of another PR ?

matttsss:
> We can merge as is. Do we wait for the other Screens in order to push a complete `MainActivity.kt` or will it be the topic of another PR ?

This is subject to another PR since it is not the aim of this fix to add a new screen

app/src/main/java/com/github/se/cyrcle/ui/navigation/NavigationActions.kt
matttsss: Yes thank you, I added the changes

Armetiis:
All good

Armetiis (approved):
All good, sorry for the accidental close

PR #14, opened 2024-10-08
Feature: Login screen

matttsss:
This PR adds a new screen to the app, allowing users to sign-in before accessing the app.


LouisJamesVasseur (approved):
Seems good to me, again quite similar in structure to Bootcamp so it works. We will definitely need to come back to this to review the UI however.

sean-prz (approved):
Testes and approved on my machine too ! 
You should probably add an How to run in the docs somewhere 
I'll let you merge

PR #15, opened 2024-10-08
cicd: add MapBox secrets to decode

andrepeiry:
Access and Downloads Tokens were added to the decode step.

Removed a duplicated step (Enable KVM group perms)


andrepeiry:
Merge with admin rights to allow quick CI upgrade to secrets.

PR #16, opened 2024-10-08
cicd: add names to secrets

andrepeiry:
Instead of simply adding the key, ./local.properties should now contains KEY=actual_key.


andrepeiry:
Use admin bypass

PR #17, opened 2024-10-08
cicd: fix typo in secret name

andrepeiry:
Secrets name were MAP_EXAMPLE instead of MAPBOX_EXAMPLE.


andrepeiry:
Use admin bypass

PR #25, opened 2024-10-10
feat: implement basic Composables

andrepeiry:
Have the basic structure for the following Atom and Molecules :
- InputText
- Toast
- BottomNavigationBar
- Button
- FloatingActionButton of different sizes

WARNING: Most of the stuff is just there as an example, and must be further implemented. Existing stuff must as well be review to conform to the graphic chart.


Pijukai (approved):
A lot of useful components that will make our code more flexible, Thanks ! 
1. Be careful not to predict too much our need , we will probably add the components when needed (`InputFileFromPhone()`).
2. A lot of lines that are not covered by tests but this will probably be addressed later since code coverage is not our priority for now (`Input.kt`).

PR #26, opened 2024-10-10
Bugfix navigation

matttsss:
This bugfix allows for navigation from the Authentication to the Map Screen as well as to the (future) list screen by adding the new BottomNavigation menu.
It also adds a test to check the presence of the map on the screen.


Pijukai:
There is a problem with the bottom navigation. When displaying the map, if we click on the map button 2 times it reloads it.

Pijukai (approved):
Everything is OK for me now. Good Job!

AlexandreBeeDeeBop (approved):
All good

PR #28, opened 2024-10-10
Implement scrollable parking spot list with clever filtering.

AlexandreBeeDeeBop:
Create a scrollable list of parking spots in the 'SpotListScreen' composable, displaying detailed information on each card. The `SpotCard` composable shows parking spot details,
 while the `FilterHeader` allows users to apply advanced
filters to sort spots based on multiple criteria such as protection, rack type, and capacity,
showing the most relevant results first.

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/list/List.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/list/List.kt
@@ -1,0 +1,449 @@
+package com.github.se.cyrcle.ui.list
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.*
+import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.foundation.lazy.LazyRow
+import androidx.compose.foundation.lazy.items
+import androidx.compose.material3.*
+import androidx.compose.runtime.*
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.shadow
+import androidx.compose.ui.graphics.Color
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/se/cyrcle/ui/list/List.kt
```diff
@@ -0,0 +1,449 @@
...
+import com.github.se.cyrcle.model.parking.*
+import com.github.se.cyrcle.ui.theme.DarkBlue
+import kotlin.math.atan2
+import kotlin.math.cos
+import kotlin.math.sin
+import kotlin.math.sqrt
+
+// EXAMPLES
+
+val referencePoint1 =
```
LouisJamesVasseur, line 26: These could be left for the test files

app/src/main/java/com/github/se/cyrcle/ui/list/List.kt
```diff
@@ -0,0 +1,449 @@
...
+            }
+      }
+}
+
+// Extension fuction to toggle selection
+fun <T> Set<T>.toggle(item: T): Set<T> {
+  return if (this.contains(item)) this - item else this + item
+}
+
+fun calculateDistance(point1: Point, point2: Point): Double {
```
LouisJamesVasseur, line 435: We'll need to modify this to include viewModel-based distance modifications, but this works just fine for now

LouisJamesVasseur (approved):
This looks all good. We'll need to get back to this to increase code coverage and connect the screen to the ViewModel. We'll also need to increase code coverage and add more tests

Pijukai (approved):
Good to go well done !

PR #29, opened 2024-10-10
CardView of the Parking Spots

LouisJamesVasseur:
The objective of this PR is to introduce a new Screen, consecutive to either Map or List View, that will allow users to see a list of a selected parking spot's attributes as a descriptive "Card": CardScreen.

Tests are provided with an appropriate code coverage.

Only the UI is available as of this PR, with dummy parking spots being used for now to display functionality. It will need to be interactive with the View Model's data via the Map/List screen.

### DIFF
--- a/app/src/androidTest/java/com/github/se/cyrcle/ui/card/CardScreenTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/ui/card/CardScreenTest.kt
@@ -2,25 +2,32 @@ package com.github.se.cyrcle.ui.card
 
 import androidx.compose.ui.test.*
 import androidx.compose.ui.test.junit4.createComposeRule
-import androidx.navigation.testing.TestNavHostController
-import androidx.test.core.app.ApplicationProvider
+import androidx.navigation.NavHostController
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.github.se.cyrcle.ui.navigation.NavigationActions
+import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
+import org.mockito.Mockito.mock
 
 @RunWith(AndroidJUnit4::class)
 class CardScreenTest {
 
   @get:Rule val composeTestRule = createComposeRule()
 
-  val mockNavController = TestNavHostController(ApplicationProvider.getApplicationContext())
-  val navigationActions = NavigationActions(navController = mockNavController)
+  private lateinit var navHostController: NavHostController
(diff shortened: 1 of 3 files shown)


app/src/androidTest/java/com/github/se/cyrcle/ui/card/CardScreenTest.kt
```diff
@@ -0,0 +1,64 @@
...
+        .onAllNodesWithTag("ParkingImage0") // First image
+        .assertCountEquals(1)
+  }
+
+  @Test
+  fun buttonsAreDisplayed() {
+    composeTestRule.setContent { CardScreen(curParking = parking1, navigationActions) }
+
+    // Verify the buttons
+    composeTestRule.onNodeWithTag("ShowInMapButton").assertExists()
```
matttsss, line 46: Better to assert that it is displayed over it exists

app/src/main/java/com/github/se/cyrcle/ui/card/CardScreen.kt
```diff
@@ -0,0 +1,289 @@
...
+        ParkingCapacity.LARGE,
+        ParkingRackType.TWO_TIER,
+        ParkingProtection.COVERED,
+        0.0,
+        true)
+
+// Main UI Composable function to display card with parking information
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun CardScreen(curParking: Parking = parking1, navigationActions: NavigationActions) {
```
matttsss, line 128: Optional argument should be placed after mandatory one

app/src/androidTest/java/com/github/se/cyrcle/ui/card/CardScreenTest.kt
```diff
@@ -0,0 +1,64 @@
...
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class CardScreenTest {
+
+  @get:Rule val composeTestRule = createComposeRule()
+
+  val mockNavController = TestNavHostController(ApplicationProvider.getApplicationContext())
```
matttsss, line 18: Overkill, could just mock it, no need to add a dependency like that yet

andrepeiry (approved):
In-code comments seem a bit excessive, most of them are not needed.
`convertXxxToString` can be obtain more easily by modifiy the `enum class Thing` into `enum class Thing(val string: String)`

matttsss (approved):
Overall good UI screen for a demo, will have to integrate the ViewModel later!

PR #38, opened 2024-10-13
fix: intermittent Authentication tests error

matttsss:
This PR intends to solve the following intermittent error in `SingInTest.kt`: "No compose hierarchies found in the app."

The suggested solution involves setting the content manually and removing an unused dependency

After 5 successful CI tests, one could maybe assume the issue is solved?


andrepeiry (approved):
Simple yet effective test modification. I hope it will "do the trick" in the long run.

LouisJamesVasseur (approved):
The fix you suggest seems reasonable and efficient, as additional mocking will probably allow the test to have more deterministic behavior. Good catch and good change!

PR #39, opened 2024-10-14
Fix: use our colors for the theme instead of default one

sean-prz:
_Closes #40_
### What
This smalls PR simplifes the theme by removing the dynamic color (Colors based on user setting/wallpaper ect).
### Why
We will not need to use dynamic color for our app, and this feature prevented our colors from being displayed.
We can now use `colors = MaterialTheme.colorScheme.primary` instead of hardcoded value


LouisJamesVasseur (approved):
This will probably make our life easier in the long run. Having Cerulean as the primary color instead of DarkBlue also makes sense given how the color scheme of our app is for now.

andrepeiry (approved):
Everything seems in order.

PR #47, opened 2024-10-15
Display markers and manage them

Pijukai:
### Content PR 

This PR enables the app to display the different parkings on the Map. It also manages how it should be displayed and when. It also handles Clustering of markers during de-zooming. Finally, it clamps the zooming so that the user cannot load the entire database by de-zooming or zoom to a scale that is not functional.   

### Why ?  

This an essential feature to our map that will allow the users to easily locate parkings.  

### How ? 

- The markers are displayed using the `AnnotationManager` instance and `AnnotationOptions` refers to a single marker. 
- The zoom is clamped using `cameraBoundsOptions`  value and then by applying it to the current map instance through its `mapView`.
- Clustering is done through `ClusterOptions`

### Testing

This is the required snippet that needs to be added in `MainActivity.kt` to be able to test. 

```kotlin
val firebase = ParkingRepositoryFirestore(Firebase.firestore)
    val firebase2 = ImageRepositoryCloudStorage(auth)
    val parkingViewModel = ParkingViewModel(firebase2, firebase)
``` 

Also update the instance of Map in `MainActivity.kt` by adding the `parkingViewModel` argument

- The zoom clamping is tested in `MapScreenTest.kt` but you can still test this through real-time utilization.
- To test the display of the markers you can uncomment the 3 parkings in `MapScreen.kt` and comment the snippet below (Pipelining parkings from the database is not yet finished). The test can be done by using the app ourselves since it is impossible to test it through `MapScreenTest.kt`.   

```kotlin
// Get parkings in the current view
                parkingViewModel.getParkingsInRect(bottomLeftCorner, topRightCorner)

                // Create PointAnnotationOptions for each parking
                val pointAnnotationOptions =
                    listOfParkings.value.map { parking ->
                      PointAnnotationOptions()
                          .withPoint(
                              Point.fromLngLat(
                                  parking.location.center.longitude(),
                                  parking.location.center.latitude()))
                          .withIconImage(resizedBitmap)
                    }
```

### Code Coverage
image

### Feature Preview

image


Pijukai:
For the reviewers, there is still a big performance issue. you need to dezoom at max the map wait for 20sec and then scroll a bit to see that the pings are displaying. Fixing performance and refactoring will be the subject of another PR.

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -44,7 +77,124 @@ fun MapScreen(navigationActions: NavigationActions) {
         MapboxMap(
             Modifier.fillMaxSize().padding(padding).testTag("MapScreen"),
             mapViewportState = mapViewportState,
-            style = { MapStyle("mapbox://styles/seanprz/cm27wh9ff00jl01r21jz3hcb1") })
+            style = { MapStyle("mapbox://styles/seanprz/cm27wh9ff00jl01r21jz3hcb1") }) {
+              DisposableMapEffect { mapView ->
+
+                // Set camera bounds options
+                val cameraBoundsOptions =
+                    CameraBoundsOptions.Builder().minZoom(8.0).maxZoom(18.0).build()
```
Armetiis, line 85: You created const values for min zoom and maxzoom but didn't use them here

Pijukai: Fixed !

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -44,7 +77,124 @@ fun MapScreen(navigationActions: NavigationActions) {
...
+                          ScreenCoordinate(
+                              centerPixel.x + viewportWidth / 2,
+                              centerPixel.y - viewportHeight / 2))
+
+                  bottomLeftCorner =
+                      mapView.mapboxMap.coordinateForPixel(
+                          ScreenCoordinate(
+                              centerPixel.x - viewportWidth / 2,
+                              centerPixel.y + viewportHeight / 2))
+
```
Armetiis, line 168: The above code could be obtained from a helper function that takes as argument a mapView and returns a tuple of points (topRight, bottomLeft), to avoid duplication in the changeListener, which does the exact same thing

Armetiis: This modularization will be particularly useful if in a future sprint you need to change the way you compute the corners

Armetiis (changes requested):
Great job on the map. Clamping the zoom level was a good idea, and the clustering is too. For a future PR, it would be nice to try and create more parking instances to see the clustering in action!  I also understand that the "debouncing" of the resquests will be done in a future PR this Sprint, as this is crucial for the functionality of the map

Please take a look at my comment regarding modularization.

Apart from that, very good job on this PR. As soon as the comment is addressed, you will get my approval

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Pijukai: I plan to do a second PR in this Sprint to improve modularization and performance.  It will aim to refactor some line of codes and handle the request to the database accordingly

Pijukai: Fixed !

Armetiis (approved):

matttsss (approved):
Nice feature, will be looking forwards for your second PR!

PR #51, opened 2024-10-16
Navigation for the process of adding new parking spots

sean-prz:
_Closes #45_
### (What)  This PR provides placeholders screens for adding new parking spots, and implement the navigation between them.

image

### (How) Key changes 
- The navigation graph defined inside the navHost has been moved to its own file, making the Main cleaner and the navGraph reusable for testing purposes.
- This PR adds for the first time the viewModel to the App

#### Notes
- This PR is not about the UI.
- The mechanism to select location and add attributes is not defined yet.
- 80% cc is reached
image

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/CyrcleNavHost.kt
+++ b/app/src/main/java/com/github/se/cyrcle/CyrcleNavHost.kt
@@ -1,0 +1,50 @@
+import androidx.compose.runtime.Composable
+import androidx.navigation.NavHostController
+import androidx.navigation.compose.NavHost
+import androidx.navigation.compose.composable
+import androidx.navigation.navigation
+import com.github.se.cyrcle.model.parking.ParkingViewModel
+import com.github.se.cyrcle.ui.add.AttributesPicker
+import com.github.se.cyrcle.ui.add.LocationPicker
+import com.github.se.cyrcle.ui.authentication.SignInScreen
+import com.github.se.cyrcle.ui.card.CardScreen
+import com.github.se.cyrcle.ui.list.SpotListScreen
+import com.github.se.cyrcle.ui.map.MapScreen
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/github/se/cyrcle/CyrcleNavHost.kt
```diff
@@ -0,0 +1,39 @@
...
+      composable(Screen.AUTH) { SignInScreen(navigationActions) }
+    }
+    navigation(startDestination = Screen.LIST, route = Route.LIST) {
+      composable(Screen.LIST) { SpotListScreen(navigationActions, parkingViewModel) }
+      composable(Screen.CARD) { CardScreen(navigationActions, parkingViewModel) }
+    }
+    navigation(startDestination = Screen.MAP, route = Route.MAP) {
+      composable(Screen.MAP) { MapScreen(navigationActions) }
+    }
+    navigation(startDestination = Screen.LOCATION_PICKER, route = Route.ADD_SPOTS) {
```
Pijukai, line 34: I think that each route should start from a TLD , here it would `Map` . This ensures that the navigation flow stay logical and consistent as a graph. Now , it is a comment about semantic and it functionally works so i won't request changes.

Pijukai (approved):
Good Job overall on this PR just don't forget to rebase so that the navigation stays updated. Good job !

app/src/main/java/com/github/se/cyrcle/CyrcleNavHost.kt
sean-prz: To be discussed with the team, I don't think that is how Route works, otherwise all our screen that have nothing in common appart being accessible from the MapScreen would be in the same group ?

Armetiis (approved):
Nothing to add

PR #43, opened 2024-10-14
General Theme improvement, with improved Atom and Molecule

andrepeiry:
Remove not implemented elements for code coverage. Expand theme to have a wider range of available colors. Update Atom and Molecule to have better function signature, as well as default values.

- Remove not-yet implemented Atom and Molecule (code coverage).
- Improve Theme by assigning our values instead of default ones.
- Simplify and expand Atom and Molecule signatures by allowing `Modifier` as a argument, as well as adding default values, for the top/bottom bars for example.

Code coverage of this PR
image

Close #39 and #30.

PR #53, opened 2024-10-16
Test coverageee

matttsss:
### What is this PR?
This PR aims to boost the Code Coverage of the project and implement the End2End test.

### How?
By improving tests on the `cyrcle.ui.list` package and the End2End test.
The End2End test currently runs with the database **online** since no feasible work-around to isolating Firestore was found during this Sprint. One major solution could be Dependency Injection to standardize the tests that use Firestore and this could be a subject for Sprint 3.

For compatibility and cohesion amongst all screens, some test tags were also added/modified.

Also, a *Sign in as guest* button was also added to the `LoginScreen` in order to bypass the `GoogleSignInIntent` during the End2End test.

### Why?
For app stability and Milestone grades 👍

### Test coverage:
Current test coverage is assumed to go up after @andrepeiry 's PR (#43) on atoms and molecules and should then reach above 80% code coverage.
image

### DIFF
--- a/app/src/androidTest/java/com/github/se/cyrcle/MainActivityTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/MainActivityTest.kt
@@ -1,0 +1,18 @@
+package com.github.se.cyrcle
+
+import androidx.compose.ui.test.junit4.createAndroidComposeRule
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class MainActivityTest {
+
+  @get:Rule val composeTestRule = createAndroidComposeRule<MainActivity>()
+
+  @Before fun setUp() {}
+
+  @Test fun reviewCardDisplaysWell() {}
+}
--- a/app/src/androidTest/java/com/github/se/cyrcle/MainActivityTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/MainActivityTest.kt
@@ -83,6 +83,7 @@ class MainActivityTest {
 
     @OptIn(ExperimentalTestApi::class)
     fun toCard(index: Int) {
+      composeTestRule.waitUntilAtLeastOneExists(hasTestTag("SpotListItem"))
       composeTestRule.onNodeWithTag("SpotListColumn").performScrollToIndex(index)
       composeTestRule
           .onNodeWithTag("SpotListColumn")
--- a/app/src/androidTest/java/com/github/se/cyrcle/MainActivityTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/MainActivityTest.kt
@@ -1,7 +1,25 @@
 package com.github.se.cyrcle
 
+import android.util.Log
+import androidx.compose.ui.test.ExperimentalTestApi
+import androidx.compose.ui.test.assertHasClickAction
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.assertTextContains
+import androidx.compose.ui.test.assertTextEquals
+import androidx.compose.ui.test.hasTestTag
+import androidx.compose.ui.test.junit4.ComposeTestRule
 import androidx.compose.ui.test.junit4.createAndroidComposeRule
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.performClick
 import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.github.se.cyrcle.model.parking.ParkingRepositoryFirestore
+import com.github.se.cyrcle.model.parking.TestInstancesParking
+import com.github.se.cyrcle.ui.navigation.TopLevelDestinations
+import com.google.firebase.Firebase
+import com.google.firebase.auth.FirebaseAuth
+import com.google.firebase.firestore.firestore
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.tasks.await
 import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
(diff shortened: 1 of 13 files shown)


sean-prz:
Thanks Mattéo for all the tests.

The button you added to sign-in doesn't render well on my emulator 

The button you added 
image

matttsss:
@sean-prz I used the default values for color given by @andrepeiry :
`getButtonColors(ColorScheme.SECONDARY)` it should get fixed in his PR

app/src/androidTest/java/com/github/se/cyrcle/MainActivityTest.kt
```diff
@@ -0,0 +1,163 @@
...
+
+  private class ListScreenRobot(val composeTestRule: ComposeTestRule) {
+
+    @OptIn(ExperimentalTestApi::class)
+    fun toCard(index: Int) {
+      composeTestRule.onNodeWithTag("SpotListColumn").performScrollToIndex(index)
+      composeTestRule
+          .onNodeWithTag("SpotListColumn")
+          .onChildAt(index + 1)
+          .assertIsDisplayed()
```
sean-prz, line 90: this assertion fails on my end for some reason. Let's double check with @andrepeiry to see if it's just me

sean-prz (changes requested):
I'm playing the devil's advocate on the comment, but you did a really nice job on the test Suite. 

I just want to check why the test doesn't pass and why you removed the cache (even if I can guess) before approving

This is quite a long PR, with a lot packed in only one issue (that you can maximum spend 4 hours on). For the future I think you should try breaking down your work into smaller issue, i.e here you could have made one PR for the ui change even if they are really minor, or one PR where you refactor old test, and one where you add more tests, ect...

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
```diff
@@ -2,35 +2,57 @@ package com.github.se.cyrcle.ui.list
...
 import com.github.se.cyrcle.ui.theme.molecules.BottomNavigationBar
-import com.mapbox.geojson.Point
 import com.mapbox.turf.TurfMeasurement
 
 @Composable
 fun SpotListScreen(
     navigationActions: NavigationActions,
     parkingViewModel: ParkingViewModel = viewModel(factory = ParkingViewModel.Factory),
 ) {
-  // TODO: Use the user's location instead of a hardcoded reference point
```
sean-prz, line 31: I would suggest to add this as an issue or has a draft issue so that we don't forget to do it

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
```diff
@@ -39,7 +61,10 @@ fun SpotListScreen(
   val sortedFilteredParkingSpots =
       parkingSpots
           .map { parking ->
-            parking to TurfMeasurement.distance(referencePoint, parking.location.center)
+            // TODO: Use the user's location instead of a hardcoded reference point
```
sean-prz, line 64: I would suggest to add this as an issue or has a draft issue so that we don't forget to do it

app/src/main/java/com/github/se/cyrcle/ui/map/overlay/ZoomControls.kt
```diff
@@ -26,17 +25,16 @@ fun ZoomControls(onZoomIn: () -> Unit, onZoomOut: () -> Unit) {
         modifier =
             Modifier.padding(15.dp)
                 .background(Color.Black.copy(alpha = 0.70f), shape = RoundedCornerShape(8.dp))
-                .sizeIn(maxWidth = 35.dp)) {
-          Box(modifier = Modifier.testTag("ZoomControlsIn")) {
```
sean-prz, line 30: I would have appreciated if you mentioned the UI changes in your PR description too. And even better if you did a small PR just for that, (Remember one task per PR)

app/src/androidTest/java/com/github/se/cyrcle/MainActivityTest.kt
```diff
@@ -0,0 +1,163 @@
...
+      composeTestRule.onNodeWithTag("RowProtectionPrice").assertIsDisplayed()
+      composeTestRule.onNodeWithTag("RowSecurity").assertIsDisplayed()
+      composeTestRule.onNodeWithTag("ButtonsColumn").assertIsDisplayed()
+      composeTestRule.onNodeWithTag("ShowInMapButton").assertIsDisplayed().assertHasClickAction()
+      composeTestRule.onNodeWithTag("AddReviewButton").assertIsDisplayed().assertHasClickAction()
+      composeTestRule.onNodeWithTag("ReportButton").assertIsDisplayed().assertHasClickAction()
+    }
+
+    @OptIn(ExperimentalTestApi::class)
+    fun goBack() {
```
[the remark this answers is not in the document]
matttsss: Each `*Robot` class exposes manipulations that humans could do, while writing the tests I figured that that function could be useful to write new tests.

app/src/androidTest/java/com/github/se/cyrcle/MainActivityTest.kt
```diff
@@ -0,0 +1,163 @@
...
+    fun assertListScreen() {
+      composeTestRule.onNodeWithTag("BottomNavigationBar").assertIsDisplayed()
+      composeTestRule.onNodeWithTag("SpotListColumn").assertIsDisplayed()
+      composeTestRule.onNodeWithTag("ShowFiltersButton").assertIsDisplayed().assertHasClickAction()
+
+      // TODO check that the cards are displayed
+    }
+
+    @OptIn(ExperimentalTestApi::class)
+    fun toMap() {
```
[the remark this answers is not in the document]
matttsss: See reply above

app/src/androidTest/java/com/github/se/cyrcle/model/parking/ParkingRepositoryFirestoreTest.kt
```diff
@@ -26,13 +23,6 @@ class ParkingRepositoryFirestoreTest {
     FirebaseApp.initializeApp(context)
 
     val db = FirebaseFirestore.getInstance()
-    val settings = firestoreSettings {
```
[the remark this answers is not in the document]
matttsss: These lines were causing issues with an error like: `Firebase error: cannot set settings more than once`.
I guess the underlying reason is that `MainActivityTest.kt` runs first and since it initializes our app, it sets the cache settings for the whole test run.

app/src/androidTest/java/com/github/se/cyrcle/MainActivityTest.kt
matttsss: Could you maybe add a `Thread.sleep(10000)` and check on the emulator screen what the screen looks like at that point?

app/src/main/java/com/github/se/cyrcle/ui/map/overlay/ZoomControls.kt
matttsss: Sorry I didn't really consider it as UI since the change didn't affect it... :sweat_smile:

sean-prz: I see the reasoning

sean-prz (approved):
The problem with the test failing on my machine has been fixed, full steam ahead!

PR #56, opened 2024-10-16
Optimize map and implement Firestore caching

### LINKED ISSUE #55: Optimize map and implement firestore caching

Armetiis:
Closes #55 
### Key changes
- Modified getScreenCorners (previously getScreenCoordinateFromPoint) function to include an optional buffer parameter.
- Updated the camera change listener to use buffered coordinates for loading parkings and unbuffered coordinates for checking bounds.
- This approach reduces unnecessary API calls while ensuring new parkings are loaded when needed.
- Implemented Firestore caching in the main activity to improve data retrieval performance.

### Coverage
image

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
@@ -2,6 +2,7 @@ package com.github.se.cyrcle.ui.map
 
 import android.graphics.Bitmap
 import android.graphics.BitmapFactory
+import android.util.Log
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
@@ -13,7 +14,7 @@ import androidx.compose.material3.Scaffold
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.collectAsState
-import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.mutableDoubleStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalContext
@@ -51,7 +52,7 @@ const val minZoom = 8.0
 fun MapScreen(
     navigationActions: NavigationActions,
     parkingViewModel: ParkingViewModel,
-    state: MutableState<Double> = remember { mutableStateOf(defaultZoom) }
+    state: MutableState<Double> = remember { mutableDoubleStateOf(defaultZoom) }
 ) {
 
   val listOfParkings = parkingViewModel.rectParkings.collectAsState()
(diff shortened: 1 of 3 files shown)


matttsss (approved):
Nice hack!

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -187,27 +178,49 @@ fun MapScreen(
       }
 }
 
+fun inBounds(
```
Pijukai, line 181: I think there is a missing documentation here.

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -110,48 +110,39 @@ fun MapScreen(
...
-                  pointAnnotationOptions.forEach { annotationManager.create(it) }
+                  // Get the top right and bottom left coordinates of the current view only when
+                  // what the user sees is outside the screen
+                  val (currentBottomLeft, currentTopRight) =
+                      getScreenCorners(mapView, useBuffer = false)
+                  if (!inBounds(
+                      currentBottomLeft, currentTopRight, loadedBottomLeft, loadedTopRight)) {
+                    Log.d("MapScreen", "Loading parkings in new view")
+                    // Get the buffered coordinates for loading parkings
+                    loadedBottomLeft = getScreenCorners(mapView, useBuffer = true).first
```
Pijukai, line 128: It's an open question but wouldn't it be better to just compute once the function `getScreenCorners()` instead of 2 times here ?

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -187,27 +178,49 @@ fun MapScreen(
...
  *
- * @param mapView The MapView
- * @return A pair of the top right and bottom left coordinates
+ * @param mapView the MapView to get the screen corners from
+ * @param useBuffer whether to use a buffer to get the corners
+ * @return a pair of the bottom left and top right corners of the screen
  */
-fun getScreenCoordinateFromPoint(mapView: MapView): Pair<Point, Point> {
-
+fun getScreenCorners(mapView: MapView, useBuffer: Boolean = true): Pair<Point, Point> {
```
Pijukai, line 203: We can make those functions private because they will be useful only for the `MapScreen`

Pijukai (changes requested):
When you adressed all the remarks you are good to go for me. That's a really good job !

Pijukai (approved):
Good job!

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Armetiis: done

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Armetiis: you are entirely right, fixed

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Armetiis: fixed

PR #61, opened 2024-10-17
Implement user data class and repository interface.

AlexandreBeeDeeBop:
### What is this PR?

This PR introduces a User data class containing all useful informations about a biker on our app. It also introduces a UserRepository interface to manage user-related data operations efficiently for my next PR.

### How?
User data class encapsulates user information such as ID, name, email, favorite parking spots and more.
UserRepository interface defines methods for user data management, enabling future database interactions.

### Why?
To establish a structured foundation for user operations, facilitating easier integration with storage solutions and improving code maintainability.

### Additional details?
A bunch of additional information is planned to be added to the user's data class and a few functions to the interface too, but not yet because of the complications they induce and the lack of time I have for this sprint. They are not essential and it can wait another 1-2 sprints. This PR does not include testing, I just need your feedback on the features I created. They cant break anything on the main. My second PR, to come later, will include testing using 2 mock users and the interface.

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/model/user/User.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/user/User.kt
@@ -1,7 +1,5 @@
 package com.github.se.cyrcle.model.user
 
-import com.github.se.cyrcle.model.parking.Location
-import com.google.firebase.Timestamp
 data class User(
     val userId: String = "",
     val username: String = "",
@@ -11,11 +9,11 @@ data class User(
     val profilePictureUrl: String? = null,
     val homeAddress: String? = null,
     val favoriteParkingSpots: List<String> = emptyList(),
-    //TODO For some reason these break my serialization/
+    // TODO For some reason these break my serialization/
     // deserialization and i dont have the time to fix it :
-    //val lastKnownLocation: Location? = null,
-    //val lastLoginTime: Timestamp? = null,
-    //val accountCreationDate: Timestamp? = null,
+    // val lastKnownLocation: Location? = null,
+    // val lastLoginTime: Timestamp? = null,
+    // val accountCreationDate: Timestamp? = null,
     val numberOfContributedSpots: Int = 0,
     val userReputationScore: Int = 0
 )
--- a/app/src/main/java/com/github/se/cyrcle/model/user/User.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/user/User.kt
@@ -1,7 +1,7 @@
 package com.github.se.cyrcle.model.user
 
 data class User(
-    val userId: String = "",
+    val userId: String,
     val username: String = "",
     val firstName: String? = null,
     val lastName: String? = null,
--- a/app/src/main/java/com/github/se/cyrcle/model/user/User.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/user/User.kt
@@ -3,14 +3,13 @@ package com.github.se.cyrcle.model.user
 data class User(
     val userId: String,
     val username: String = "",
-    val firstName: String? = null,
-    val lastName: String? = null,
+    val firstName: String = "",
+    val lastName: String = "",
     val email: String = "",
-    val profilePictureUrl: String? = null,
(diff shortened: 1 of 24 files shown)


AlexandreBeeDeeBop:
oops i forgot ktfmt format thingy

app/src/main/java/com/github/se/cyrcle/model/user/User.kt
```diff
@@ -0,0 +1,19 @@
...
+    val lastName: String? = null,
+    val email: String = "",
+    val profilePictureUrl: String? = null,
+    val homeAddress: String? = null,
+    val favoriteParkingSpots: List<String> = emptyList(),
+    // TODO For some reason these break my serialization/
+    // deserialization and i dont have the time to fix it :
+    // val lastKnownLocation: Location? = null,
+    // val lastLoginTime: Timestamp? = null,
+    // val accountCreationDate: Timestamp? = null,
```
LouisJamesVasseur, line 16: Good to have commented these fields for now. We'll have to look into why this happens, but the fields aren't super relevant yet so it's not too big of an issue as of right now

app/src/main/java/com/github/se/cyrcle/model/user/User.kt
LouisJamesVasseur, line 1: We'll have to look into which we make mandatory for the user to input (for eg maybe a profile picture isn't mandatory?)

app/src/main/java/com/github/se/cyrcle/model/user/User.kt
```diff
@@ -0,0 +1,19 @@
+package com.github.se.cyrcle.model.user
+
+data class User(
+    val userId: String = "",
```
sean-prz, line 4: I do not see why we allow a default value for userId to be the empty string? shouldn't we forbid the creation of User Object if no userId is specified ?

LouisJamesVasseur: Agree with this, but I'm not sure it's too much of an issue at this point in time. As I said, we'll need to figure out what fields are mandatory or not but this is a design decision for a meeting.

app/src/main/java/com/github/se/cyrcle/model/user/UserRepository.kt
LouisJamesVasseur, line 1: Everything here seems up to standard for a ViewModel, and well commented

LouisJamesVasseur (approved):

app/src/main/java/com/github/se/cyrcle/model/user/User.kt
matttsss: I kind of agree with Sean, this is kind of crucial for testing

Pijukai: I strongly agree as well.

LouisJamesVasseur (changes requested):
I think that the userID default value should be fixed ASAP, but the rest looks good

LouisJamesVasseur (approved):
As the userID issue has been fixed, I'll approve this PR. Thanks for the quick fix!

app/src/main/java/com/github/se/cyrcle/model/user/User.kt
```diff
@@ -0,0 +1,19 @@
+package com.github.se.cyrcle.model.user
+
+data class User(
+    val userId: String,
+    val username: String = "",
+    val firstName: String? = null,
```
Pijukai, line 6: I think we should replace `String? = null` by `String = ""` for the other inputs for consistency. We do not expect any other type besides `String` and except for the username they can be empty if the user doesn't want to give corresponding info.

Pijukai (changes requested):
Address all the comments and I approve.

Pijukai (approved):
All good just pass the CI and you can merge

AlexandreBeeDeeBop:
bruh i already did idk why it didnt work, my bad

LouisJamesVasseur (approved):
Everything seems good to merge!

PR #62, opened 2024-10-17
Patch to prevent query density 

Pijukai:
## PR content

This PR is a small patch that aims to prevent making too much queries of parking while using the map. 

## Why ? 

While using a `CameraListener` to update the markers of the map, we found out that the map makes too much queries (one batch of query for each pixel displacement). This was addressed in a previous PR. This PR address the issue of the rotation motion which is a edge case. 

## How ? 

I disabled the `rotateGestureDetector` so that the map doesn't detect rotating movement from the user. I also make the buffer bigger as a safety measure. 

## Testing

This can be tested by using the map and try to rotate it.


matttsss (approved):
All good for me :ok_hand:

Armetiis (approved):
The fix seems good

PR #64, opened 2024-10-17
Review Screen

LouisJamesVasseur:
### What?

This PR aims to implement a Review Screen to add reviews. Please note that this screen is not yet integrated with the ViewModel from my previous PR as of this Sprint. I have also edited some things with regards to navigation and the view model that are in relation w/ the screen in questions

### Why?

This will allow users to add reviews on parkings, allowing users to know which parking spots are good quality on our app.

### Testing
90% code coverage has been achieved. The screen is quite basic therefore so are the tests

### What it should look like
Screenshot 2024-10-17 at 20 39 51

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/review/ReviewScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/review/ReviewScreen.kt
@@ -1,0 +1,125 @@
+package com.github.se.cyrcle.ui.review
+
+import android.widget.Toast
+import androidx.compose.foundation.layout.*
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Star
+import androidx.compose.material.icons.filled.StarBorder
+import androidx.compose.material.icons.filled.StarHalf
+import androidx.compose.material3.*
+import androidx.compose.runtime.*
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import com.github.se.cyrcle.model.parking.ParkingViewModel
+import com.github.se.cyrcle.model.review.ReviewViewModel
+import com.github.se.cyrcle.ui.navigation.NavigationActions
+import com.github.se.cyrcle.ui.navigation.Screen
+import com.github.se.cyrcle.ui.theme.molecules.TopAppBar
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun ReviewScreen(
+    navigationActions: NavigationActions,
+    parkingViewModel: ParkingViewModel,
+    reviewViewModel: ReviewViewModel
+) {
+  var sliderValue by remember { mutableStateOf(0f) }
+  var textValue by remember { mutableStateOf("") }
+
+  val context = LocalContext.current // Get the current context
+
+  Scaffold(
+      topBar = {
+        TopAppBar(
+            navigationActions = navigationActions,
+            title = {
+              Text(
+                  text = "Add Your Review",
+                  fontSize = 20.sp,
+                  fontWeight = FontWeight.Bold,
+                  color = MaterialTheme.colorScheme.secondary,
+                  modifier = Modifier.testTag("TopAppBarTitle"))
+            })
+      }) { paddingValues ->
+        Column(
+            modifier = Modifier.fillMaxSize().padding(paddingValues).padding(16.dp),
+            horizontalAlignment = Alignment.CenterHorizontally,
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/github/se/cyrcle/ui/review/ReviewScreen.kt
```diff
@@ -0,0 +1,125 @@
...
+                  }
+
+              // Display the experience text based on the slider value
+              Text(
+                  text =
+                      when (sliderValue) {
+                        in 0f..1f -> "Terrible experience!"
+                        in 1f..2f -> "Bad experience :("
+                        in 2f..3f -> "Poor experience :/"
+                        3f -> "Average experience..."
```
matttsss, line 79: Do you not have the `in 3f..3.5f` notation here? I don't know how this notation works 😅

LouisJamesVasseur: It met my expectations on UI so I didn't change it yet. This is probably a UI detail best discussed in a meeting?

AlexandreBeeDeeBop (commented):
Maybe we should prevent reviews from being empty or even too short ? Like make the button not clickable until

app/src/main/java/com/github/se/cyrcle/ui/review/ReviewScreen.kt
```diff
@@ -0,0 +1,125 @@
...
+                        in 3.5f..4f -> "Good experience!"
+                        in 4.5f..5f -> "Great experience!"
+                        else -> ""
+                      },
```
andrepeiry, line 83: I think that having non-linear opinions (i.e. the interval is not the same for each line) is bad for the user (cognitive charge)

andrepeiry, line 83: Just a note for later, but this `String` will have to be put in a `.xml` file

andrepeiry, line 83: I'm not 100% onboard with the smileys, especially as there are not consistent along all lines (some have 1, some have none, and the `...` after `Average experience` might convey a "sarcasm" vibes)

LouisJamesVasseur: I agree with this actually! Was not sure where to draw the line as "Average" and UI-wise it didn't seem too jarring so I didn't change it yet

LouisJamesVasseur: I'd like to integrate some actual emojis down the line (I figured it'd create some sort of visual cue to homogenize which "feeling" maps to what score) , so these are just meant to be placeholders. I'll note this though, I'll bring up the general UI of this screen at the scrum meeting

app/src/main/java/com/github/se/cyrcle/ui/review/ReviewScreen.kt
```diff
@@ -0,0 +1,125 @@
...
+                  label = { Text("Write your review") },
+                  modifier = Modifier.fillMaxWidth().testTag("ReviewInput"))
+
+              Spacer(modifier = Modifier.height(16.dp))
+
+              // Add Review Button
+              Button(
+                  onClick = {
+                    Toast.makeText(context, "Review Added!", Toast.LENGTH_SHORT).show()
+                    navigationActions.navigateTo(Screen.CARD)
```
matttsss, line 111: This should be a `navigationActions.goBack()` otherwise the go back of the CardScreen leads back to the ReviewScreen (the navigation stack is not poped)

andrepeiry (changes requested):
Lots of nice code in there. I have a few problems with the texts displayed during the selection of Stars quality, most of them being the lack of linearity (blocking for me).
In general, comments are welcome, but some feels very pedantic and worsen the readability.

app/src/main/java/com/github/se/cyrcle/ui/review/ReviewScreen.kt
LouisJamesVasseur: Fixed in commit!

app/src/main/java/com/github/se/cyrcle/ui/review/ReviewScreen.kt
andrepeiry: I would say that either of the following scheme would be valid
- 1, 1.5, 2, ... : but the texts might need to be more confusing than helpful
- 1-2, 2.5-3.5, 4-5 : Three steps, might be too simple
- 1-1.5, 2-2.5, ... : Five steps, might be the best

LouisJamesVasseur: I like this a lot, this will be a future design choice we'll implement in a future sprint

andrepeiry (approved):

matttsss (approved):
Looks all good for me 👌

PR #74, opened 2024-10-18
Fixes the Release workflow

sean-prz:
_Closes #75_
This PR fixes the building of the APK by building in debug mode which does the signature for us.
It also changes this action to a manually triggered one, so that we can name the version and give it a code for the apk

Tested as you can see below : 

image
image


andrepeiry (approved):
Small note about missing a blank line at the end, but the code seems in good condition.

matttsss (approved):

PR #88, opened 2024-10-26
Refactor the (de)serialization using an adapter

Armetiis:
This PR simplifies and improves the serialization and deserialization of `Parking` objects, focusing on the proper handling of `Point` objects within the `Location` structure.

### Changes
- Removed the `Point2D` helper class, directly working with Mapbox `Point` objects.
- Simplified `serializeParking` and `deserializeParking` methods using the new `PointAdapter`.

### Benefits
- Cleaner, more maintainable code with reduced complexity.
- Improved performance by eliminating multiple conversions between different point representations.

### Testing
- Verified that serialization produces the correct format for Firestore storage.
- Ensured deserialization correctly reconstructs `Parking` objects from serialized data.
- Confirmed that queries based on geographical data still function as expected.


sean-prz (approved):
Very satisfying !

I tested adding new spots to the db via the UI to see make sure the serialization and its inverse wasn't broken by this PR.

 Everything is smooth.

PR #90, opened 2024-10-26
Feature: Dependency injection

### LINKED ISSUE #63: Stabilise E2E tests
Implement dependency injection to avoid reliance on Firebase for the tests.

matttsss:
### What is this PR?
This PR aims to reduce dependency of tests on Firebase by using Hilt for dependency injections.

### How?
Hilt was added to the project framework and injects repositories into the `MainActivity`.
It currently has three modules:
* `AppModule`: This module provides the `FirebaseFirestore` and `FirebaseAuth` instances with the correct settings
*  `RepoModule`: This is the main module used to provide production repositories
*  `MockRepoModule`: This is the testing module used to provide Mock repositories during testing

This PR also provides Mock implementations for the repositories and can be found in `app/src/androidTest/java/com/github/se/cyrcle/di/mocks`, to be used when necessary in future tests.

### Code Coverage:
image

Closes #63

### DIFF
--- a/app/src/androidTest/java/com/github/se/cyrcle/di/mocks/MockParkingRepository.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/di/mocks/MockParkingRepository.kt
@@ -1,0 +1,101 @@
+package com.github.se.cyrcle.di.mocks
+
+import com.github.se.cyrcle.model.parking.Parking
+import com.github.se.cyrcle.model.parking.ParkingRepository
+import com.github.se.cyrcle.model.parking.TestInstancesParking
+import com.mapbox.geojson.Point
+import com.mapbox.turf.TurfMeasurement
+import javax.inject.Inject
+
+class MockParkingRepository @Inject constructor() : ParkingRepository {
+  private var uid = 0
+  val parkings = mutableListOf<Parking>(TestInstancesParking.parking1)
+
+  override fun getNewUid(): String {
+    return (uid++).toString()
+  }
+
+  override fun onSignIn(onSuccess: () -> Unit) {
+    onSuccess()
+  }
+
+  override fun getParkings(onSuccess: (List<Parking>) -> Unit, onFailure: (Exception) -> Unit) {
+    if (parkings.none { it.uid == "" }) onSuccess(parkings)
+    else onFailure(Exception("Error getting parkings"))
+  }
+
+  override fun getParkingById(
+      id: String,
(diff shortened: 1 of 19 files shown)


Armetiis (approved):
Thanks chief 🫡

More seriously, amazing work you did here to integrate Dagger Hilt to the project. The mock folder with the provided mock repositories is a great idea.

Also, thank you for aligning the build.gradle.kts!

app/src/androidTest/java/com/github/se/cyrcle/di/mocks/MockParkingRepository.kt
```diff
@@ -0,0 +1,101 @@
...
+import com.github.se.cyrcle.model.parking.Parking
+import com.github.se.cyrcle.model.parking.ParkingRepository
+import com.github.se.cyrcle.model.parking.TestInstancesParking
+import com.mapbox.geojson.Point
+import com.mapbox.turf.TurfMeasurement
+import javax.inject.Inject
+
+class MockParkingRepository @Inject constructor() : ParkingRepository {
+  private var uid = 0
+  val parkings = mutableListOf<Parking>(TestInstancesParking.parking1)
```
sean-prz, line 12: Everything in the mocks are private expect for this val

sean-prz (approved):
Wow this is a huge amount of works to understand Hilt and then write explicitly every function of the repository !
Thank you so much, it will make testing way more pleasant.

tests still works on my end, and the app functionality is preserved. We're good to go.

app/src/androidTest/java/com/github/se/cyrcle/di/mocks/MockParkingRepository.kt
matttsss: Right, that was for debuging the E2E tests xD

sean-prz:
We will still need to address the fact that we can't add more parking spot to the db without braking the unit test for the parking repository. 

I don't think that Dependency Injection can fix this as we want to test the actual repository, what did you think of Maria's idea of adding the debug variable to the repo to change collection path

Armetiis (approved):

matttsss:
@sean-prz I do not think that this is a good idea, we would still rely on Firebase and keep on making requests. I realy do not think  that our repo tests should exist in the `androidTest` folder. I beleive that Firestore should be mocked just like it currently is in the `ParkingRepositoryFirestoreTest` from the `test` folder

PR #91, opened 2024-10-26
Implement MapViewModel

sean-prz:
_Closes #77_

### 1. This PR adds a viewModel for the Map.
The view models states are : 
- The camera position (state). with two setters, one normal and one that reset the bearing and pitch.
Used to remember where the user was on the map, when he navigates between screens. So he is put back where he left off.

- The selected Location, with setter/getter
Stores the new parking spots location selected by the user to pass it on to the next screen to set the attributes.

- screenCoordinates, with  setter/getter
To pass the screenCoordinates of the selected Rectangles to a higher composable that can get use them to convert them into coordinates

- locationPickerState with setter/getter
Stores an enum that reflects at which steps of the process of adding a new parking spot the user i.

### 2. This PR adds an MapConfig object
This is to avoid code duplication has the two screens that both use a map share a lot of their configs. 
This then introduce functions for the following configs settings 
- Set the default camera position 
- Set out map styles
- Set the camera position the be equal to the one stored in the map viewmodel. 

### Notes
This PR keeps the CC above 80% by providing a suite of unit tests for the viewmodel.
All of this is already used in code and ready to be merged, but the implementation will come in a later PR to break them down into smaller ones.

image


app/src/test/java/com/github/se/cyrcle/model/map/mapViewModelTest.kt
```diff
@@ -0,0 +1,74 @@
...
+import com.github.se.cyrcle.model.parking.Location
+import com.mapbox.geojson.Point
+import com.mapbox.maps.CameraState
+import com.mapbox.maps.EdgeInsets
+import com.mapbox.maps.ScreenCoordinate
+import org.junit.Test
+
+class MapViewModelTest {
+
+  @SuppressLint("StateFlowValueCalledInComposition")
```
Armetiis, line 14: why this annotation? I don't think it applies anywhere in this class

app/src/test/java/com/github/se/cyrcle/model/map/mapViewModelTest.kt
```diff
@@ -0,0 +1,74 @@
...
+  @Test
+  fun mapViewModelTestCameraState() {
+    val mapViewModel = MapViewModel()
```
Armetiis, line 17: you can instantiate a single viewModel for the whole class rather than one per function

Armetiis (commented):
Great work. Very useful for consistency across screens and modulatization.
A small remark: the file names (mapViewModel and mapViewModelTest) are in lowercase when they should be in uppercase

Apart from that and the two comments I left you, this is perfect.

Armetiis (approved):

Pijukai (approved):
That's good work and i really like the concept behind the rectangles. I think we should discuss about integrating markers so that the users will be comfortable since similar apps use markers. That will be the topic of a future meeting and will probably be a task of a next sprint. 
Good Job !

PR #92, opened 2024-10-26
Adding Parking :  The Location Picker Screen

sean-prz:
_Closes #72_

### Choosing the location of the new parking (What)

This PR implements the first of two screens necessary to add a new parking spot.
Specifically : 

- The user first set the top left corner by aligning it with the crosshair and pressing the next button
- The user set the bottom right corner by dragging the screen to change the rectangle dimension
- The user can toggle the map movements with a button on/off if he needs to adjust his map.

### How

**Rectangle** 

The rectangle is drawn on top of the map on a Canvas, and is detached from the map. 
Once the rectangle is set, with the help of the viewmodel, we can retrieve in the mapview the screen coordinates to convert them to a Location Object, and save it before going on. 
This implementation makes the map freeze during the selection of the bottom right corner, hence the button to toggle movements again.

**LocationPickerState**
The state in the viewmodel of this name is used to determine in which steps of the selection the user is, and the ui elements are determined by this state that is being observed by every composable

**Button**
Atoms were not used as they are not ready at this stage.

**Tests**
As a reviewer you can ignore the tests, they are hard to understand and badly made, I'll refactor them using Hilt on next sprint

image
image

image


app/src/main/java/com/github/se/cyrcle/ui/add/location/LocationPickerTopBar.kt
```diff
@@ -0,0 +1,55 @@
...
+            Column {
+              Text(
+                  "Where is the Parking ?",
+                  color = MaterialTheme.colorScheme.primary,
+                  fontSize = MaterialTheme.typography.headlineLarge.fontSize,
+                  fontFamily = MaterialTheme.typography.bodyLarge.fontFamily,
+                  modifier = Modifier.padding(start = 16.dp),
+                  fontWeight = FontWeight.Bold)
+              if (locationPickerState == LocationPickerState.NONE_SET) {
+                Text(
```
LouisJamesVasseur, line 43: I feel like the text isn't clear enough. "Set the top-left corner of the Parking" should be accompanied with a mention of how the user should actually do that ("Set the top-left corner of the Parking using crosshair below" would be enough)

app/src/main/java/com/github/se/cyrcle/ui/add/location/LocationPickerTopBar.kt
```diff
@@ -0,0 +1,55 @@
...
+                    "Set the top-left corner of the parking",
+                    color = MaterialTheme.colorScheme.tertiary)
+              } else if (locationPickerState == LocationPickerState.TOP_LEFT_SET) {
+                Text(
+                    "Set the bottom-right corner of the parking",
```
LouisJamesVasseur, line 48: I don't feel like this text is accurate enough either. The point of this screen is to create a rectangle, so I don't feel like this accurately describes the requested task. The fact that you drag (this is the central UI action for this screen) should be mentioned somewhere

LouisJamesVasseur (approved):
I've checked the UI behavior of the screens and it looks good to me (besides the texts not being accurate enough in my opinion as I mention in my comments). I don't see any immediate issues w/ the code either, though I think some more documentation/comments would be nice (even if code is very clear so not really an issue).

app/src/main/java/com/github/se/cyrcle/ui/add/location/LocationPickerTopBar.kt
sean-prz: Good remark, I'll update the text before merging

sean-prz:
> I've checked the UI behavior of the screens and it looks good to me (besides the texts not being accurate enough in my opinion as I mention in my comments). I don't see any immediate issues w/ the code either, though I think some more documentation/comments would be nice (even if code is very clear so not really an issue).

Yeah I should have added more documentation, sorry. Will make an effort on the next PRs

app/src/main/java/com/github/se/cyrcle/ui/add/location/LocationPicker.kt
```diff
@@ -0,0 +1,80 @@
...
+          DisposableMapEffect { mapViewInstance ->
+            mapView.value = mapViewInstance
+            onDispose {}
+          }
+        }
+        RectangleSelection(mapViewModel, padding)
+        Crosshair(mapViewModel, padding)
+      }
+
+  LaunchedEffect(locationPickerState) {
```
LouisJamesVasseur, line 55: I'm not sure it's extremely useful to disable the gestures on the map. I think the rectangle should take care of capturing the inputs?

sean-prz: Oh yeah my bad this is some old code from a previous implementation, I have to remove it

app/src/main/java/com/github/se/cyrcle/ui/add/location/overlay/RectangleSelection.kt
```diff
@@ -0,0 +1,92 @@
...
+          horizontalArrangement = Arrangement.Center,
+          modifier = Modifier.fillMaxSize(),
+      ) {
+        androidx.compose.material3.Button(
+            onClick = { mapGesturesEnabled.value = !mapGesturesEnabled.value },
+            content = { Text("Toggle Rotation") })
+      }
+    }
+  }
+  if (locationPickerState == MapViewModel.LocationPickerState.BOTTOM_RIGHT_SET) {
```
Armetiis, line 80: After manually testing the process, I think it would be more flexible to allow users not only to toggle the rotation but rather to be able to lock and unlock all types of movements (sliding the map, zooming AND rotating). This would greatly inprove the user experience.

sean-prz: Sure, This can easily be changed and is a great improvement !

Armetiis (approved):
You did a really good job. This is very intuitive and the implementation is very clever!

PR #93, opened 2024-10-26
Adding Parking : The Attributes Picker Screen

sean-prz:
_closes #73_

#### This PR adds the screen to select the attributes.
The user can after having selected the location now set the attributes, and then click submit, the parking spot is then added to the database and the user redirected to the map.

#### How
The core of the feature is in the package `add.attributes` the rest is just here to adapt the project to this new screen

This PR reuse the drawRectangle function of the map screen that draws the rectangles to show it as a preview in the topBar.
It also adds two customs Composables to make the DropDown Menus and radios buttons that make the core of this AttributesPicker Screen. 

image
<br>
image

#### Notes : 
- The default title will be replaced with the auto name suggestor implemented by Maria
- Don't pay attention to the test, they will be remade with Hilt more properly on next sprint


andrepeiry (changes requested):
[Blocking] A lot of composable that are in your code could be placed inside `/ui/theme/{atom, molecules}`, mostly the `TextInputField`s and similar objects.
[Note] Your particular `BottomBarAddAttr` could be generalized and moved into `atom, molecules` as well. But that can be done in another PR.

[Blocking] Even thought it is not this PR content (but related as this is the continuum of another), I think `add` is too generic for a package name and doesn't tell anything about its content. It could be improved as `addParking` for example.

[Note] I saw `hasCCTV` in a file. The guideline doesn't specify anything about name convention, but it is good practice to have acronyms also in CamelCase (so the name should be `hasCctv`). In general, it is better to avoid acronym and use whole words, such as `hasVideoSurveillance`

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
```diff
@@ -117,6 +117,10 @@ class ParkingViewModel(
         { Log.e("ParkingViewModel", "Error getting parkings: $it") })
   }
 
+  fun getUid(): String {
```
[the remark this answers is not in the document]
sean-prz: True

app/src/main/java/com/github/se/cyrcle/ui/add/attributes/AttributesPicker.kt
```diff
@@ -0,0 +1,258 @@
...
+                  label = "How many bikes can we park ?",
+              )
+              DynamicSelectTextField(
+                  options = ParkingRackType.entries.toList(),
+                  selectedValue = rackType,
+                  label = "Which type of rack is in the parking spot?",
+              )
+              Spacer(modifier = Modifier.height(16.dp))
+              BooleanRadioButton(
+                  question = "Is the parking spots under camera surveillance ?", hasSecurity)
```
AlexandreBeeDeeBop, line 124: Minor typo

sean-prz:
> [Blocking] A lot of composable that are in your code could be placed inside `/ui/theme/{atom, molecules}`, mostly the `TextInputField`s and similar objects. [Note] Your particular `BottomBarAddAttr` could be generalized and moved into `atom, molecules` as well. But that can be done in another PR.
> 
> [Blocking] Even thought it is not this PR content (but related as this is the continuum of another), I think `add` is too generic for a package name and doesn't tell anything about its content. It could be improved as `addParking` for example.
> 
> [Note] I saw `hasCCTV` in a file. The guideline doesn't specify anything about name convention, but it is good practice to have acronyms also in CamelCase (so the name should be `hasCctv`). In general, it is better to avoid acronym and use whole words, such as `hasVideoSurveillance`

Thanks for the great review ! Those are all very relevant point, I'm gonna fix them and come back to you

app/src/main/java/com/github/se/cyrcle/ui/add/attributes/AttributesPicker.kt
sean-prz: Thanks, Fixed with lastest commit

andrepeiry (approved):
Most of the work has been done, I'll make some minute changes in my PR about the integration of the Atom & Molecules inside the code [not yet created, insert link later], so thank you for your quick work !

PR #99, opened 2024-10-29
Feature implement interaction with markers

Pijukai:
## Content of the PR 

This PR enables interaction with the markers. On click , the camera move to put the selected marker in the center of the Screen and a preview of the marker's content is displayed.  

## Why ? 

This allows the user to be informed quickly of important information about the selected parking without having to go to CardScreen.  This also enhance the interaction between the user and the map improving the number of things that the user can do just with one click. 

## How ? 

The preview card is defined in `layout/item_callout_view.xml`. Then it is displayed in MapScreen using a `ViewAnnotationManager`. There is a lot of listeners that are being used in the code. They aim to implement the expected behavior displayed to the user and to prevent any unexpected behavior that could be linked to the fact that MapBox works concurrently on multiple threads and this needs to be handled. 

## Additional resources
image

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
@@ -4,6 +4,7 @@ import android.annotation.SuppressLint
 import android.graphics.Bitmap
 import android.graphics.BitmapFactory
 import android.util.Log
+import android.view.View
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
@@ -78,7 +79,7 @@ fun MapScreen(
     }
   }
 
-  val cancelables = remember<List<Cancelable>> { emptyList() }.toMutableList()
+  var cancelables = remember<List<Cancelable>> { emptyList() }.toMutableList()
 
   val context = LocalContext.current
 
@@ -138,8 +139,15 @@ fun MapScreen(
                       textNativeView.text = "Custom text here"
                     }
                   }
-                  cancelables += mapView.mapboxMap.subscribeMapIdle(listener)
+                  cancelables.forEach(Cancelable::cancel)
+                  cancelables = listOf(mapView.mapboxMap.subscribeMapIdle(listener)).toMutableList()
                   true
+                })
+
+            mapView.setOnClickListener(
+                View.OnClickListener {
+                  cancelables.forEach(Cancelable::cancel)
+                  viewAnnotationManager.removeAllViewAnnotations()
                 })
 
             var (loadedBottomLeft, loadedTopRight) = getScreenCorners(mapView, useBuffer = true)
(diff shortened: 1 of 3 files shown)


Pijukai:
About testing, i don't think it's possible to click on markers on MapBox map since it relies on OpenGL rendering and they are not native components of UI. Hence the test coverage is about 79%. I will ask the coach about it so you can review the PR in the mean time.

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -39,15 +44,22 @@ import com.mapbox.maps.plugin.annotation.AnnotationConfig
...
+import com.mapbox.maps.viewannotation.annotationAnchor
+import com.mapbox.maps.viewannotation.geometry
+import com.mapbox.maps.viewannotation.viewAnnotationOptions
 
 const val defaultZoom = 16.0
 const val maxZoom = 18.0
 const val minZoom = 8.0
+const val LAYER_ID = "0128"
 
+@SuppressLint("ResourceType", "SetTextI18n")
```
sean-prz, line 62: The annotation isn't useful on my end. Removing it didn't add any warning to the code and it ran the same

sean-prz (approved):
All good for me, tested it and it works as intended.


I will start a new branch off your PR to start working on some issues that we have with the mapScreen but they are unrelated to this PR. I hope it will help with coverage

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -89,7 +106,50 @@ fun MapScreen(
...
+                            AnnotationSourceOptions(clusterOptions = ClusterOptions()),
+                        layerId = LAYER_ID))
+
+            annotationManager.addClickListener(
+                OnPointAnnotationClickListener {
+                  removeViewAnnotation = false
+                  viewAnnotationManager.removeAllViewAnnotations()
+
+                  // recenter the camera on the marker if it is not the case already
+                  if (mapViewportState.cameraState?.center != it.point) {
```
sean-prz, line 118: useless to check if it's not the center, 99% it won't be the case, And if it is already the center, it changes nothing to center it again

sean-prz (approved):

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
LouisJamesVasseur: I agree with this, though I'm not sure it impacts performance much as is.

LouisJamesVasseur (approved):
I've tested the behavior and it behaves as expected with good code. Nice job!

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Pijukai: fixed!

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Pijukai: Fixed !

sean-prz (approved):

LouisJamesVasseur (approved):
Even after all the new changes, everything still works well so I'm re-approving this pull request!

PR #111, opened 2024-10-29
User Profile Screen First Version

AlexandreBeeDeeBop:
Content of the PR
This PR adds the user's profile screen where he will be able to see his personal information such as name, username, profile picture, etc. but also his list of favorite parking spots ! The user is also able to modify some of this data .

Why ?
The user might want to see his information, or change his information, or simply have a quick look at his favorite parking spots.

How ?
I added a bunch of buttons, text, and a few images by using their URL. They are clickable and send you to different interfaces. There is an editing mode but I will need to work on this more later so that it actually edits stuff once you click on save changes. For the list of favorite spots there is a clickable yellow-filled star which turns empty to symbolize the parking spot being removed from favorites.

Testing
Pretty trivial tests where I check that everything is shown correctly, everything that i want to click is clickable.

### DIFF
--- a/app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ProfileScreenTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ProfileScreenTest.kt
@@ -2,29 +2,29 @@ package com.github.se.cyrcle.ui.profile
 
 import androidx.compose.ui.test.*
 import androidx.compose.ui.test.junit4.createComposeRule
-import androidx.navigation.compose.rememberNavController
+import androidx.navigation.NavHostController
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.github.se.cyrcle.ui.navigation.NavigationActions
+import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
+import org.mockito.Mockito.mock
 
 @RunWith(AndroidJUnit4::class)
 class ProfileScreenTest {
   @get:Rule val composeTestRule = createComposeRule()
+  private lateinit var mockNavController: NavHostController
 
-  private fun setupProfileScreen() {
-    composeTestRule.setContent {
-      val navController = rememberNavController()
-      val navigationActions = NavigationActions(navController)
-      ProfileScreen(navigationActions = navigationActions)
-    }
+  @Before
+  fun setupProfileScreen() {
+    mockNavController = mock(NavHostController::class.java)
+    val navigationActions = NavigationActions(mockNavController)
+    composeTestRule.setContent { ProfileScreen(navigationActions = navigationActions) }
   }
 
   @Test
   fun testInitialDisplayMode() {
-    setupProfileScreen()
-
     composeTestRule.waitUntil(timeoutMillis = 5000) {
       composeTestRule.onAllNodesWithTag("ProfileImage").fetchSemanticsNodes().isNotEmpty()
     }
@@ -40,8 +40,6 @@ class ProfileScreenTest {
 
   @Test
   fun testEditModeTransition() {
-    setupProfileScreen()
-
     composeTestRule.waitUntil(timeoutMillis = 5000) {
       composeTestRule.onAllNodesWithTag("EditButton").fetchSemanticsNodes().isNotEmpty()
     }
@@ -63,8 +61,6 @@ class ProfileScreenTest {
 
   @Test
   fun testCancelEditMode() {
-    setupProfileScreen()
-
     composeTestRule.waitUntil(timeoutMillis = 5000) {
       composeTestRule.onAllNodesWithTag("EditButton").fetchSemanticsNodes().isNotEmpty()
     }
@@ -102,8 +98,6 @@ class ProfileScreenTest {
 
   @Test
   fun testSaveChanges() {
-    setupProfileScreen()
-
     composeTestRule.waitUntil(timeoutMillis = 5000) {
       composeTestRule.onAllNodesWithTag("EditButton").fetchSemanticsNodes().isNotEmpty()
     }
@@ -135,8 +129,6 @@ class ProfileScreenTest {
 
   @Test
   fun testFavoriteParkingsSection() {
-    setupProfileScreen()
-
     composeTestRule.waitUntil(timeoutMillis = 5000) {
       composeTestRule.onAllNodesWithTag("FavoriteParkingsTitle").fetchSemanticsNodes().isNotEmpty()
     }
@@ -150,8 +142,6 @@ class ProfileScreenTest {
 
   @Test
   fun testProfileImageInteraction() {
-    setupProfileScreen()
-
     composeTestRule.waitUntil(timeoutMillis = 5000) {
       composeTestRule.onAllNodesWithTag("ProfileImage").fetchSemanticsNodes().isNotEmpty()
     }
--- a/app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ProfileScreenTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ProfileScreenTest.kt
@@ -12,7 +12,6 @@ import org.junit.runner.RunWith
 @RunWith(AndroidJUnit4::class)
 class ProfileScreenTest {
 
-
   @get:Rule val composeTestRule = createComposeRule()
 
   @Test
--- a/app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ProfileScreenTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ProfileScreenTest.kt
@@ -11,163 +11,162 @@ import org.junit.runner.RunWith
 
 @RunWith(AndroidJUnit4::class)
 class ProfileScreenTest {
-    @get:Rule
-    val composeTestRule = createComposeRule()
+  @get:Rule val composeTestRule = createComposeRule()
 
-    private fun setupProfileScreen() {
-        composeTestRule.setContent {
-            val navController = rememberNavController()
-            val navigationActions = NavigationActions(navController)
-            ProfileScreen(navigationActions = navigationActions)
-        }
+  private fun setupProfileScreen() {
+    composeTestRule.setContent {
+      val navController = rememberNavController()
+      val navigationActions = NavigationActions(navController)
+      ProfileScreen(navigationActions = navigationActions)
+    }
+  }
+
+  @Test
+  fun testInitialDisplayMode() {
+    setupProfileScreen()
+
+    composeTestRule.waitUntil(timeoutMillis = 5000) {
+      composeTestRule.onAllNodesWithTag("ProfileImage").fetchSemanticsNodes().isNotEmpty()
     }
 
-    @Test
-    fun testInitialDisplayMode() {
-        setupProfileScreen()
+    // Verify initial display mode elements
+    composeTestRule.onNodeWithTag("ProfileImage").assertExists()
+    composeTestRule.onNodeWithTag("EditButton").assertExists()
+    composeTestRule.onNodeWithTag("DisplayFirstName").assertExists()
+    composeTestRule.onNodeWithTag("DisplayLastName").assertExists()
+    composeTestRule.onNodeWithTag("DisplayUsername").assertExists()
+    composeTestRule.onNodeWithTag("FavoriteParkingsTitle").assertExists()
+  }
 
-        composeTestRule.waitUntil(timeoutMillis = 5000) {
-            composeTestRule.onAllNodesWithTag("ProfileImage").fetchSemanticsNodes().isNotEmpty()
-        }
+  @Test
+  fun testEditModeTransition() {
+    setupProfileScreen()
 
-        // Verify initial display mode elements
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/se/cyrcle/ui/profile/ProfileScreen.kt
```diff
@@ -0,0 +1,192 @@
...
+
+  var favoriteParkings by remember { mutableStateOf(user.favoriteParkings.toMutableList()) }
+  val parkingFavorites = remember {
+    mutableStateListOf(*List(user.favoriteParkings.size) { true }.toTypedArray())
+  }
+
+  var profilePictureUrl by remember { mutableStateOf(user.profilePictureUrl) }
+  val defaultImageUrl =
+      "https://th.bing.com/th/id/R.aee6adef085a77dfa4708f3fd4a1ffb5?rik=nj%2buNqgLIJU0JQ&pid=ImgRaw&r=0"
+  val filledStarImageUrl =
```
Armetiis, line 41: You shouldn't use images for this. Even less fetch them online at every single composition. Please use material 3 provided icons

app/src/main/java/com/github/se/cyrcle/ui/profile/ProfileScreen.kt
```diff
@@ -0,0 +1,192 @@
...
+  var lastName by remember { mutableStateOf(TextFieldValue(user.lastName)) }
+  var username by remember { mutableStateOf(TextFieldValue(user.username)) }
+
+  var favoriteParkings by remember { mutableStateOf(user.favoriteParkings.toMutableList()) }
+  val parkingFavorites = remember {
+    mutableStateListOf(*List(user.favoriteParkings.size) { true }.toTypedArray())
+  }
+
+  var profilePictureUrl by remember { mutableStateOf(user.profilePictureUrl) }
+  val defaultImageUrl =
```
Armetiis, line 39: Instead of fetching from online every time this screen is composed, add as a ressource to the project

app/src/main/java/com/github/se/cyrcle/ui/profile/ProfileScreen.kt
```diff
@@ -0,0 +1,192 @@
...
+
+@Composable
+fun ProfileScreen(user: User) {
+  var isEditing by remember { mutableStateOf(false) }
+  var firstName by remember { mutableStateOf(TextFieldValue(user.firstName)) }
+  var lastName by remember { mutableStateOf(TextFieldValue(user.lastName)) }
+  var username by remember { mutableStateOf(TextFieldValue(user.username)) }
+
+  var favoriteParkings by remember { mutableStateOf(user.favoriteParkings.toMutableList()) }
+  val parkingFavorites = remember {
```
Armetiis, line 34: It is hard to tell what this is and what it does. Is this just for demo purposes? I think if we click on the star, and therefore remove a parking from favorites, it should simply dissapear (opt: maybe we can add a confirmation card in a future sprint).
You don't need to "simulate" this feature while the viewmodel is not integrated. It seems like unnecesary overhead

app/src/main/java/com/github/se/cyrcle/ui/profile/ProfileScreen.kt
```diff
@@ -0,0 +1,192 @@
...
+                value = username,
+                onValueChange = { username = it },
+                label = { Text("Username") },
+                modifier = Modifier.testTag("UsernameField"))
+
+            val painter =
+                rememberAsyncImagePainter(
+                    ImageRequest.Builder(context)
+                        .data(
+                            if (profilePictureUrl.isNotBlank()) profilePictureUrl
```
Armetiis, line 85: Detail: You can use this pretty nice syntax: 
`profilePictureUrl.ifBlank { defaultImageUrl })`

app/src/main/java/com/github/se/cyrcle/ui/profile/ProfileScreen.kt
Armetiis: You could use a material 3 icon for this purpose too, giving us more liberty in shapes, colors, etc.

Armetiis (changes requested):
Please take a look at my comments first before reading the following block.

After making the changes suggested in the comments, we could get rid of the use of async images, making unnecessary internet requests. The only time if would indeed be needed is if the user does have a profile picture

matttsss (changes requested):
To add to what Maria said:
* Your file `ProfileScreen.kt` file is missing it's package name
* You can "factorise" the card composable from the `LazyRow` into it's own `Composable` function like you did in the `ListScreen`
* How can you navigate in and out of your screen?
* You can also use Andre's InputText` Composable from the  `atoms` package instead of the `TextFields`

app/src/main/java/com/github/se/cyrcle/ui/profile/ProfileScreen.kt
```diff
@@ -0,0 +1,192 @@
...
+  val emptyStarImageUrl =
+      "https://cdn0.iconfinder.com/data/icons/thin-voting-awards/57/thin-232_star_favorite_add-1024.png"
+
+  val context = LocalContext.current
+  val imagePickerLauncher =
+      rememberLauncherForActivityResult(ActivityResultContracts.GetContent()) { uri ->
+        uri?.let { profilePictureUrl = it.toString() }
+      }
+
+  DisposableEffect(Unit) {
```
matttsss, line 52: I'm not sure what this is meant to do, how does it affect the UI?

andrepeiry:
Just a comment on the Git messages : please read the guidelines !
- The commit description doesn't start with a capital letter
- There is no scope `style`. The one to use is `refactor`
- Avoid commit description longer than 50 characters
- I think you spotted it, but your commit with the command inside could have been avoided before push.

What you can do to improve your commit history in AndroidStudio :
- Open your Git panel (bottom left)
- Left click on your commit, and change the commit message to better follow the guidelines
- To remove the two styling commits (and make a simple correct one) :
  - Reset (soft) your branch to the last good commit (in this case : `test: Add comprehensive UI tests for ProfileScreen`, `8abb3a7`
  - Commit your changes (with, for example `refactor: format with KTfmt`)
- Try to push your change. If it gives you an error, force push (`git push -f origin YOUR_BRANCH`)

sean-prz:
Morning,  can you also add a picture of what the screen is supposed to look like so everyone that is not a reviewer can see your work without checking out your branch, thanks

Armetiis:
I would like to raise an important point, that I failed to mention in my review
I might be wrong on this, feel free to argue your point of view:

I think it would be much easier and "cleaner" to have 2 separate screens:
- ProfileOverviewScreen: would correspond to "isEditing == false"
- ProfileEditScreen: would correspond to "isEditing == true"

This would simplify the logic so much, getting rid of so many "if statements" (e.g. to display the save changes and cancel buttons, or the modify button)

One might argue that this would imply code duplication. True, but I think the maintanability and simplicity of the code is a priority.

matttsss:
To add on what Maria said, you could focus your first PR on simply a ProfileScreen with the UserViewModel (and no editing) and the second on the EditProfileScreen for example, it will clearer and more maintainable for later.

LouisJamesVasseur:
As a reply to @Armetiis's comment.
I agree, to some extent. I'd like to contribute a UI/UX perspective rather than a code one (from a pure code point of view, Maria is definitely correct). 
I'm of the opinion that having multiple screens for the same feature isn't an amazing idea as it just creates confusion. The question is whether editing and viewing the profile are separate features? In the current implementation, I would argue not. Given that we're only adapting 3 text fields for now, I think the way @AlexandreBeeDeeBop did it for now makes the experience quite seamless and cohesive, but this might be put into question if we expand the editing capabilities of a profile.
Overall, I'd say this is a design choice best suited for a discussion during a meeting

app/src/main/java/com/github/se/cyrcle/ui/navigation/NavigationActions.kt
Armetiis, line 1: Why did you delete the documentation for this file?

app/src/main/java/com/github/se/cyrcle/ui/navigation/NavigationActions.kt
```diff
@@ -22,74 +24,44 @@ object Screen {
...
       TopLevelDestination(route = Route.MAP, icon = Icons.Outlined.LocationOn, textId = Route.MAP)
+  val PROFILE =
+      TopLevelDestination(
+          route = Route.PROFILE, icon = Icons.Outlined.Person, textId = Route.PROFILE)
 }
 
-/** List of top level destinations in the app. */
-val LIST_TOP_LEVEL_DESTINATION = listOf(TopLevelDestinations.MAP, TopLevelDestinations.LIST)
+val LIST_TOP_LEVEL_DESTINATION =
+    listOf(TopLevelDestinations.MAP, TopLevelDestinations.LIST, TopLevelDestinations.PROFILE)
```
Armetiis, line 43: (Not very important rn) I don't think profile should be a TLD. For now we should keep it like this, we can discuss this with the rest of the team another time.

app/src/main/java/com/github/se/cyrcle/ui/profile/ProfileScreen.kt
```diff
@@ -1,192 +1,295 @@
...
-      "https://th.bing.com/th/id/R.aee6adef085a77dfa4708f3fd4a1ffb5?rik=nj%2buNqgLIJU0JQ&pid=ImgRaw&r=0"
-  val filledStarImageUrl =
-      "https://th.bing.com/th/id/R.f493b7a899f0c38194622d00ca70a7ff?rik=MvAc4%2fM5MeISxA&pid=ImgRaw&r=0"
-  val emptyStarImageUrl =
-      "https://cdn0.iconfinder.com/data/icons/thin-voting-awards/57/thin-232_star_favorite_add-1024.png"
+  var firstName by remember { mutableStateOf("John") }
+  var lastName by remember { mutableStateOf("Doe") }
+  var username by remember { mutableStateOf("johndoe") }
+  var profilePictureUrl by remember { mutableStateOf("") }
+  var favoriteParkings by remember { mutableStateOf(ArrayList<String>()) }
```
Armetiis, line 44: It would be preferable to use a general list rather than arrayList. This also implies changing some other lines like 238, 239 and other places where ArrayList is specified.

app/src/main/java/com/github/se/cyrcle/ui/profile/ProfileScreen.kt
```diff
@@ -1,192 +1,295 @@
...
 
   val context = LocalContext.current
```
Armetiis, line 46: I don't think the context is needed anymore, correct me if I'm wrong

Armetiis (commented):
Wow! Great work. The screen looks good and the integration of the atoms, molecules and theme make is really great! I like how you modularized the code into preview and edit composables, and how you created many composables to make it readable maintanable.
Amazing improvement!!

Please mind my comments, in particular the one about the dissapearing documentation. Thank you for addressing the small changes (ArrayList) i pointed out. Once you fix this you have my approval

app/src/main/java/com/github/se/cyrcle/ui/navigation/NavigationActions.kt
AlexandreBeeDeeBop: I think it should, it is like that on every single app I use on my phone. Debatable indeed.

app/src/main/java/com/github/se/cyrcle/ui/navigation/NavigationActions.kt
AlexandreBeeDeeBop: I asked chatgpt for something and he wiped it all, f him

AlexandreBeeDeeBop:
ProfileScreen

AlexandreBeeDeeBop:
image

app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ProfileScreenTest.kt
```diff
@@ -0,0 +1,172 @@
...
+import com.github.se.cyrcle.ui.navigation.NavigationActions
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class ProfileScreenTest {
+  @get:Rule val composeTestRule = createComposeRule()
+
+  private fun setupProfileScreen() {
```
matttsss, line 16: Here this function could be annotated with `@Before` and removed from the body of other `@Test` functions.
The `@Before` makes it that it is called automatically before every test, so that you don't have to call it in your functions.

Armetiis (approved):

app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ProfileScreenTest.kt
```diff
@@ -0,0 +1,172 @@
...
+    composeTestRule.setContent {
+      val navController = rememberNavController()
```
matttsss, line 18: Here the navController is not used and could just be mocked, there is no need to use the android one

matttsss (changes requested):
Very good improvement of your code, after those two changes, it will be all good for me!

matttsss (approved):
Nice! Thanks a lot for the changes!

Issue #104, opened 2024-10-29
Markers aren't always shown when in view of the map


sean-prz:
#### Problem

The default camera position on the map screen doesn't load the marker initially, user has to scroll off and back into the zone to see the markers

#### Solution
Use LaunchedEffect to make sure markers are displayed when on the map

Issue #103, opened 2024-10-29
Camera is reset on default position when switching between screens.


sean-prz:
#### Problem

when switching between the map screen and the add screen the camera state is not preserved, leading to a undesirable experience

#### Solution
use the mapviewmodel to store the camera state, and restore it when switching between screen

PR #113, opened 2024-10-30
refactor: refactor mapscreen

sean-prz:
_Closes #103 #104 #104 #112_
#### (What) Refactors the mapscreen
This PR refactors the mapScreen, with a bunch of small changes to fix some bugs and make it more readable, easier to maintain. The bugs and issues are all linked on top.

#### (How) Details  
In particular, this PR does the following : 
- add a LaunchedEffect onto the listOfParkings, which will update the markers whenever the list changes. This way the markers are always visible.
- renames the annotationManager into PointAnnotationManager to distinguish it from the future other annotationManager
- Save the state of the camera when pressing Add button so that user can return to the same position when switching between screens
- change deprecated cameraListener into a subscrible state
- Use MapConfig function to set style and initial position instead of hardcoded.
- rename the state named "state" with "zoomState"

#### (Why) 
The mapscreen had some undesired behaviors such as the markers not loading initially, the camera state not being preserved between screens. Furthermore the code used a deprecated feature (cameraListener) and could be made more readable and maintainable. 

The global user experience doesn't change so no need for a screenshot. Same as test coverage.

Thanks for reviewing! 

image

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
@@ -12,17 +12,21 @@ import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material3.Scaffold
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
-import androidx.compose.runtime.mutableDoubleStateOf
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.unit.dp
 import com.github.se.cyrcle.R
 import com.github.se.cyrcle.databinding.ItemCalloutViewBinding
+import com.github.se.cyrcle.model.map.MapViewModel
 import com.github.se.cyrcle.model.parking.Location
+import com.github.se.cyrcle.model.parking.Parking
 import com.github.se.cyrcle.model.parking.ParkingViewModel
 import com.github.se.cyrcle.ui.map.overlay.AddButton
 import com.github.se.cyrcle.ui.map.overlay.ZoomControls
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -66,32 +69,28 @@ const val LAYER_ID = "0128"
 fun MapScreen(
     navigationActions: NavigationActions,
     parkingViewModel: ParkingViewModel,
-    state: MutableState<Double> = remember { mutableDoubleStateOf(defaultZoom) }
```
Pijukai, line 69: I think the Ci doesn't work because you remove state which is an value created for the tests.

sean-prz: yeah yeah the PR is not marked as ready for review, this is fixed in the final version

sean-prz: It is ready now, you can actually start reviewing

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -153,69 +155,44 @@ fun MapScreen(
...
+                  }
 
-                parkingViewModel.getParkingsInRect(loadedBottomLeft, loadedTopRight)
+                  // Get the top right and bottom left coordinates of the current view only when
+                  // what the user sees is outside the screen
+                  val (currentBottomLeft, currentTopRight) =
+                      getScreenCorners(mapView, useBuffer = false)
+                  if (!inBounds(
+                      currentBottomLeft, currentTopRight, loadedBottomLeft, loadedTopRight)) {
+                    Log.d("MapScreen", "Loading parkings in new view")
```
Pijukai, line 178: Is it intended to stay or is it just for debugging sake ?

sean-prz: I don't know either, I didn't put this Log.d(), let's ask @Armetiis

Armetiis: Is it really that bad to keep some Logs? I really don't think logs should be systematically deleted. I often find myself writing the same logs over and over because I deleted them before pushing. In this case, given that we still have a problem with loading the parkings with certain map settings, I think it's better to keep this log

Pijukai (approved):
Everything is good for me and the CI passed so approved. By the way well done with the code coverage

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Pijukai: I was just asking, it's fine if it stays and as you said it can be useful. Just wanted to point it out so that we don't have useless line of code

Armetiis (approved):
Nothing to add. Good job

PR #84, opened 2024-10-23
Address Model, Repository, and ViewModel for Reverse Geocoding

Armetiis:
Closes #87 

This pull request introduces a comprehensive solution for reverse geocoding within our application, leveraging the OpenStreetMap Nominatim API. The following components have been implemented:

##### Address Data Class:
- Introduced a new Address data class to represent various address components.
- Utilized @SerializedName annotations to manage inconsistencies in API response labels (e.g., city/town and district/suburb).

##### AddressRepositoryNominatim:
- Implemented a repository using OkHttpClient to perform reverse geocoding requests via the Nominatim API.
Chose Nominatim over Mapbox given the very limited amount of free requests the latter provides

##### AddressViewModel:
Developed AddressViewModel to manage address state and facilitate searches through the repository.
The search function accepts a Point and returns an Address

### Tests:

Comprehensive unit tests have been implemented for all new components:

##### AddressTest:
- Verifies correct construction of Address objects with various parameters.
- Ensures proper functionality of the displayRelevantFields method.

##### AddressRepositoryNominatimTest:
Tests the repository's search functionality with mock HTTP responses.
Verifies correct handling of successful and failed API requests.
Checks proper URL construction and request headers.

##### AddressViewModelTest:
Confirms that the ViewModel correctly interacts with the repository.
Verifies that search results update the ViewModel's state.

### Future Work:
Integrate the address search functionality with the parking suggestion feature

### Coverage
image

PR #115, opened 2024-10-30
Integrate title suggestion in attribute picker

### LINKED ISSUE #101: Integrate addressViewModel into Attribute pickerScreen

Armetiis:
Closes #101

This PR integrates feature implemented in PR #84 into the Attribute Picker Screen.
Additional changes are trivial changes (e.g. initialize AddressViewModel and add it to navgraph)

This is what kind of title might be suggested:
image

No additional tests (feature is already tested and screen too. This PR doesn't anything that would justify new tests)

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/addParking/attributes/AttributesPicker.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/addParking/attributes/AttributesPicker.kt
@@ -18,6 +18,7 @@ import androidx.compose.material3.Scaffold
 import androidx.compose.material3.Text
 import androidx.compose.material3.VerticalDivider
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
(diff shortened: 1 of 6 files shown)


sean-prz (changes requested):
Nice work.

We should try to decrease the number of calls to the API, currently we call it 6 times for the same address.

Also when the title is too long it overflows the topBar, we can either make the fontsize smaller or truncate the title, as you want. 

image


Address these two things and we're good to go

Armetiis:
> We should try to decrease the number of calls to the API, currently we call it 6 times for the same address.

I've tried to wrap the call to search in a LaunchedEffect. I tried different implementations but whatever I did I either did not succeed to reduce the number of calls, either failed to make it work (search returned no result). Any suggestions?

app/src/main/java/com/github/se/cyrcle/ui/addParking/attributes/AttributesPicker.kt
```diff
@@ -64,19 +66,21 @@ import com.mapbox.maps.plugin.gestures.generated.GesturesSettings
...
-    mapViewModel: MapViewModel
+    mapViewModel: MapViewModel,
+    addressViewModel: AddressViewModel
 ) {
-  val location = mapViewModel.selectedLocation.collectAsState().value!!
-  val title = remember {
-    mutableStateOf(
-        "${("%.2f").format(location.center.longitude())}, ${("%.3f").format(location.center.latitude())}")
-  }
+  val title = remember { mutableStateOf("") }
```
LouisJamesVasseur, line 72: Out of curiosity, why was this changed?

LouisJamesVasseur (approved):
Very cool feature! I find it very well implemented both from a UI and a code perspective! Great job!

app/src/main/java/com/github/se/cyrcle/ui/addParking/attributes/AttributesPicker.kt
Armetiis: The initial value of the title is irrelevant, given that as soon as the titleSuggestion is loaded from the AddressViewModel, it will get replaced. The previous value was just a placeholder @sean-prz used to avoid empty titled parkings and facilitate debugging

sean-prz (approved):
Perfect, thanks for the changes

PR #116, opened 2024-10-30
General Refactor of Test Files

LouisJamesVasseur:
### What is This PR?

This PR implements a general refactor of test files for Cyrcle, so that test files are where they are supposed to be, and to extend the use of mocking for the ViewModels 

### Why?

It is essential for organization's sake that we have an appropriate organization of our files going forward as our project is expanding.


matttsss (approved):
Thank you for refactoring our tests. Since we have mocked Firestore in Unit tests, we didn't need to instantiate it any more in these Android tests

Pijukai (approved):
Thanks you for taking on the task ,it will help cleaning the project which is one of the main goals of this Sprint. Good Job!

PR #118, opened 2024-10-30
Implement Reviews into the App

LouisJamesVasseur:
### What is this PR?

This is the first of 2 PRs for our app to actually interact with the ReviewViewModel so that reviews can actually be added and handled

### Why? 

It will be essential going forward that we can actually simulate the insertion of Reviews for our app

### How ? 

The Average Review Score for a Parking spot is now visible from the Card Screen, and Adding a review saves it in Firestore

### Tests

Tests are not directly provided for this PR, as the tests for the ViewModel functionality were provided along with the ViewModel, and tests for the ReviewScreen are still relevant to test functionality with the viewModel added. Additional UI tests will be provided with my follow-up PR, which will focus on UI integration

### Code Coverage

Screenshot 2024-10-31 at 10 39 10

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/model/parking/Parking.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/parking/Parking.kt
@@ -30,7 +30,7 @@ data class Parking(
     val price: Double,
     val hasSecurity: Boolean,
     var nbReviews: Int = 0,
-    var avgScore: Double = Double.NaN
+    var avgScore: Double = 0.0
 )
 
 interface ParkingAttribute {
--- a/app/src/main/java/com/github/se/cyrcle/model/parking/Parking.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/parking/Parking.kt
@@ -30,7 +30,7 @@ data class Parking(
     val price: Double,
     val hasSecurity: Boolean,
     var nbReviews: Int = 0,
-    var avgScore: Double = 0.0
+    var avgScore: Double = 1.0
 )
 
 interface ParkingAttribute {
--- a/app/src/main/java/com/github/se/cyrcle/model/parking/Parking.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/parking/Parking.kt
@@ -15,6 +15,8 @@ import com.mapbox.geojson.Point
  * @property protection Protection of the parking spot from the weather.
  * @property price Price of the parking spot. The price is in the currency of the country where the
  *   parking spot is located. 0 if the parking spot is free.
+ * @property nbReviews Number of Reviews for this Parking spot as of now.
+ * @property avgScore Average Review Score for this Parking spot as of now.
  */
 data class Parking(
     val uid: String,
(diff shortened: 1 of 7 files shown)


Pijukai:
I know it's not up to your PR but since you modify the tests i will bring it up. I think you can modify the ReviewScreenTest so that it uses the mock repositories that matttsss did and not the one from mockito ? Or maybe i am just mistaken

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
```diff
@@ -121,6 +121,11 @@ class ParkingViewModel(
     return parkingRepository.getNewUid()
   }
 
+  fun updateReviewScore(newScore: Double, parking: Parking) {
```
Pijukai, line 124: Since the parking is selected when the user decides to review or edit , is it necessary to put the parking as argument instead of using `selectedParking`

LouisJamesVasseur:
For the mock repositories: you're absolutely right, I just pushed a commit addressing this!

app/src/main/java/com/github/se/cyrcle/ui/review/ReviewScreen.kt
```diff
@@ -114,6 +120,14 @@ fun ReviewScreen(
           Button(
               onClick = {
                 Toast.makeText(context, "Review Added!", Toast.LENGTH_SHORT).show()
+                reviewViewModel.addReview(
+                    Review(
+                        owner = "default",
+                        text = textValue,
+                        parking = selectedParking.uid,
+                        rating = sliderValue.toDouble(),
```
Pijukai, line 128: When doing a manual test i have 0.4997 instead of 0.5 which is probably caused by the cast to double. Can you double check ?

LouisJamesVasseur: Good catch, a new commit addresses this (reviews are now doubles to to 2 decimal places)

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
Pijukai: Good fix with the default argument and well done to have updated the doc to specify the condition to use the default argument. Since we will mainly use this function to add a review the default argument is good way of simplifying the use of the function by others developers .

app/src/main/java/com/github/se/cyrcle/ui/review/ReviewScreen.kt
Pijukai: I think I still have the problem.

sean-prz (changes requested):
Hey, 
except if I am missing something, the feature described in the PR doesn't seem to work on my end: 

>  "The Average Review Score for a Parking spot is now visible from the Card Screen"

 I add review to the spot, but when I come back to it later on , the rating is not taken into account in the score. 
All of the parkings have a review score of 0.0.

LouisJamesVasseur:
Thanks @sean-prz for bringing this to my attention, you tested a case that I didn't for mine, and normally the fix I just pushed should solve the issue

app/src/main/java/com/github/se/cyrcle/model/parking/Parking.kt
```diff
@@ -26,8 +28,9 @@ data class Parking(
     val rackType: ParkingRackType,
     val protection: ParkingProtection,
     val price: Double,
-    val hasSecurity: Boolean
-    // TODO: Add list of reviews when implemented
+    val hasSecurity: Boolean,
+    var nbReviews: Int = 0,
+    var avgScore: Double = 1.0
```
sean-prz, line 33: Setting the default avgScore to be 1.0 is missleading. A user will think a new spot is a badly rated one.

I think Double.NaN is a better idea, and then change the cardScreen so that if the avgScore is Nan, display something to indicate than the rating is undefined rather than 0.0 or 1.0

sean-prz (changes requested):
There is still this one thing I commented about that bothers me but after that you're good to go for me.

I just noticed the reviewmodel from last sprint has a very low code coverage, maybe you could add an issue, so that this can be fixed in Sprint 4

app/src/main/java/com/github/se/cyrcle/ui/review/ReviewScreen.kt
```diff
@@ -114,6 +122,16 @@ fun ReviewScreen(
           Button(
               onClick = {
                 Toast.makeText(context, "Review Added!", Toast.LENGTH_SHORT).show()
+                // to avoid problematic castings
+                val sliderToValue = (sliderValue * 100).toInt() / 100.0
```
sean-prz, line 126: This is never used

app/src/main/java/com/github/se/cyrcle/model/parking/Parking.kt
LouisJamesVasseur: Good call. I'll implement this right away

LouisJamesVasseur: One slight issue: NaN causes problems with serialization. Does it seem like an appropriate fix to you if I put the default value at 0.0, and ensure that the text on CardScreen displays "No Reviews Yet" if nbReviews is 0? This way the default value is never shown to users

sean-prz: Sounds reasonable 👍

sean-prz (approved):
Good job !

app/src/main/java/com/github/se/cyrcle/ui/review/ReviewScreen.kt
LouisJamesVasseur: Fixed!

app/src/main/java/com/github/se/cyrcle/ui/review/ReviewScreen.kt
Pijukai: Apparently the problem is also present on the main so this will need a bugfix of it's own and it does not exist for all devices. I will approve your PR so that you can proceed to the 2nd one but we will need to fix this next sprint.

Pijukai (approved):
Good Job !

PR #124, opened 2024-10-31
View All Reviews for a Parking

LouisJamesVasseur:
### What is this PR?

This PR aims to further integrate the ReviewViewModel into the app by allowing users to not only see the average score, but also all reviews and their content, from the Card Screen, via a new Screen: the AllReviewsScreen. Note that user integration (ie linking Reviews to Users) is still pending for a future sprint.

It should look like this:

Screenshot 2024-10-31 at 17 50 04

### Why? 

With this, users can finally have a clear overview of what reviews were provided and add them

### How? 

The new screen allows users to scroll through all the provided Reviews for a Parking Spot by clicking on the "See All Reviews" button from Card Screen

### Testing

Tests with adequate code coverage are provided for the Screen

### Code Coverage

Screenshot 2024-10-31 at 17 43 35

### DIFF
--- a/app/src/androidTest/java/com/github/se/cyrcle/ui/card/CardScreenTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/ui/card/CardScreenTest.kt
@@ -8,6 +8,8 @@ import androidx.compose.ui.test.onChildAt
 import androidx.compose.ui.test.onChildren
 import androidx.compose.ui.test.onNodeWithTag
 import androidx.test.ext.junit.runners.AndroidJUnit4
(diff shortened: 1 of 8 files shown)


andrepeiry (approved):
Everything seems in place. PR #125 will have to go over `CardScreen` and `ReviewScreen`.

AlexandreBeeDeeBop (approved):
It's ugly but it's 9pm so it's nice.

AlexandreBeeDeeBop:
Missed opportunity at ranking them from best noted to worst noted ?  Rating should be x/5.0 maybe ?

PR #126, opened 2024-10-31
User profile favorite parkings

AlexandreBeeDeeBop:
What is this PR?
This PR enhances the favorite parkings feature of the user profile screen.
It also prepares for future viewmodel implementations.

Why?
Users should be able to remove some obsolete favorite parkings, but not fear accidentally removing one they still like.

How?
I added a second step to deleting a favorite spot by asking for confirmation.
I removed the possibility to modify parking cards from the edit mode.

Additional ressources?
image

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/profile/ProfileScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/profile/ProfileScreen.kt
@@ -277,7 +277,7 @@ private fun ProfileImage(
 }
 
 @Composable
-fun FavoriteParkingsSection(
+private fun FavoriteParkingsSection(
     favoriteParkings: List<Parking>,
     onFavoritesUpdated: (List<Parking>) -> Unit
 ) {
--- a/app/src/main/java/com/github/se/cyrcle/ui/profile/ProfileScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/profile/ProfileScreen.kt
@@ -36,39 +36,52 @@ import com.github.se.cyrcle.ui.theme.molecules.BottomNavigationBar
 
 @Composable
 fun ProfileScreen(navigationActions: NavigationActions) {
-  var isEditing by remember { mutableStateOf(false) }
-  var firstName by remember { mutableStateOf("John") }
-  var lastName by remember { mutableStateOf("Doe") }
-  var username by remember { mutableStateOf("johndoe") }
-  var profilePictureUrl by remember { mutableStateOf("") }
-  var favoriteParkings by remember { mutableStateOf(emptyList<String>()) }
+    var isEditing by remember { mutableStateOf(false) }
+    var firstName by remember { mutableStateOf("John") }
+    var lastName by remember { mutableStateOf("Doe") }
+    var username by remember { mutableStateOf("johndoe") }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/se/cyrcle/ui/profile/ProfileScreen.kt
```diff
@@ -26,6 +26,12 @@ import androidx.compose.ui.unit.dp
 import coil.compose.rememberAsyncImagePainter
 import coil.request.ImageRequest
 import coil.transform.CircleCropTransformation
+import com.github.se.cyrcle.ui.card.Location
+import com.github.se.cyrcle.ui.card.Parking
```
Pijukai, line 30: The paths for those imports are weird , it should come from the model not another Screen

app/src/main/java/com/github/se/cyrcle/ui/profile/ProfileScreen.kt
```diff
@@ -41,7 +47,43 @@ fun ProfileScreen(navigationActions: NavigationActions) {
   var lastName by remember { mutableStateOf("Doe") }
   var username by remember { mutableStateOf("johndoe") }
   var profilePictureUrl by remember { mutableStateOf("") }
-  var favoriteParkings by remember { mutableStateOf(emptyList<String>()) }
+  var favoriteParkings by remember {
```
Pijukai, line 50: Try to use the parkings of `TestInstances` and not harcoded values. Those test instances are more complete and it avoids hardcoding.

app/src/main/java/com/github/se/cyrcle/ui/profile/ProfileScreen.kt
AlexandreBeeDeeBop: Done

app/src/main/java/com/github/se/cyrcle/ui/profile/ProfileScreen.kt
LouisJamesVasseur: I second this. It'll allow for more consistency across the board

AlexandreBeeDeeBop: True, they're better then the ones I came up with. Didnt know they existed, thanks

LouisJamesVasseur (approved):
I've tested UI, and it looks good. The changes to the code make sense, and will help us implement the Users to the app better for the next Sprint. Nice Job!

app/src/main/java/com/github/se/cyrcle/ui/profile/ProfileScreen.kt
```diff
@@ -271,9 +277,9 @@ private fun ProfileImage(
 }
 
 @Composable
-private fun FavoriteParkingsSection(
-    favoriteParkings: List<String>,
-    onFavoritesUpdated: (List<String>) -> Unit
+fun FavoriteParkingsSection(
```
Pijukai, line 280: Maybe you should let the function private since it will be only useful for this screen.

matttsss:
Hey!  Just to mention, it would be nice to have a screenshot of the code coverage :)

Pijukai (approved):
Well done ! good to go thanks for the fixes

PR #132, opened 2024-11-01
Sign the APK

sean-prz:
_Closes #79_

#### What
This PR makes the manual release workflow produce a signed APK.

#### Why 
The current workflow produces a debug APK, that prevents google sign-in.
No google sign-in removes all the community part of the app. 
(Signing the APK is also required to publish the app on the play store)

##### How
This PR follows  the [official guide](https://developer.android.com/build/building-cmdline#gradle_signing) to sign your APK,  but adds encryption and use secrets to hide our signature files as this repo is public. 

#### Test
This was tested on the mock cyrcle repo of the org with success. 
I will still double check after this PR is merged that everything also work on this repo.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -55,9 +55,11 @@ jobs:
           LOCAL_PROPERTIES: ${{ secrets.LOCAL_PROPERTIES }}
           MAPBOX_ACCESS_TOKEN: ${{ secrets.MAPBOX_ACCESS_TOKEN }}
           MAPBOX_DOWNLOADS_TOKEN: ${{ secrets.MAPBOX_DOWNLOADS_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/CI.yaml
andrepeiry, line 1: Do we need the secret on the CI ? If they are only needed for the release, why bother adding them here ?

sean-prz: The gradle assemble fails if it doesn't see the keystore file and password mentionned in the build.gradle . So I added them to this CI too, there is probably other ways, but adding the secret to this CI does the trick.

matttsss (approved):
Looks nice! Good idea swapping it to a release version

andrepeiry (approved):
Everything seems in order ! Thank you for the precision in answering why the secrets where also decoded in `CI.yaml`.

PR #134, opened 2024-11-02
BugFix Preview Card don't display after adding a review

### LINKED ISSUE #131: Bugfix : when adding review, preview Card on markers don't display
When adding a review the preview card is not displaying on the map anymore when selecting a marker.

Pijukai:
Close #131 

## What is the content of the PR

This PR fixes the following bug : 
When a user add a review , comes back to the Map and select a marker , the preview card isn't displaying anymore. 

## How ? 

The problem came from the val `listener` inside the clickListener. Adding a review seems to influence the listeners inside the `MapScreen`. The fix was to define `listener` as a var outside other listeners.

edit: the problem seems to come from the fact that the map doesn't update when a parking is modified(review is added). So we need to add `selectedParking.nbreviews` to the arguments of the `LaunchedEffect` which will redraw the markers and update the `PointAnnotationManager` when the number of review changes. 

## Testing

Since it is Map-related i advise you to test manually the map by checking preview cards are displaying, adding a review then checking that the preview card still display. 

## Code Coverage 
image

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
@@ -81,6 +81,7 @@ fun MapScreen(
   val mapViewportState = MapConfig.createMapViewPortStateFromViewModel(mapViewModel)
   var removeViewAnnotation = remember { true }
   var cancelables = remember<Cancelable> { Cancelable({}) }
+  var listener: MapIdleCallback? = null
   var pointAnnotationManager by remember { mutableStateOf<PointAnnotationManager?>(null) }
 
   val bitmap = BitmapFactory.decodeResource(LocalContext.current.resources, R.drawable.red_marker)
@@ -133,7 +134,9 @@ fun MapScreen(
 
                   val pointAnnotation = it
 
-                  val listener = MapIdleCallback {
+                  listener = null
+
+                  listener = MapIdleCallback {
 
                     // Add the new view annotation
                     val viewAnnotation =
@@ -164,7 +167,7 @@ fun MapScreen(
                     removeViewAnnotation = true
                   }
                   cancelables.cancel()
-                  cancelables = mapView.mapboxMap.subscribeMapIdle(listener)
+                  cancelables = mapView.mapboxMap.subscribeMapIdle(listener!!)
                   true
                 })
 
--- a/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
@@ -83,11 +83,12 @@ fun MapScreen(
   var cancelables = remember<Cancelable> { Cancelable({}) }
   var listener = remember<MapIdleCallback?> { null }
   var pointAnnotationManager by remember { mutableStateOf<PointAnnotationManager?>(null) }
+  val selectedParking = parkingViewModel.selectedParking.collectAsState().value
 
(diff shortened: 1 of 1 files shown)


LouisJamesVasseur (changes requested):
I've tested and the issue still isn't fixed on my end...

Pijukai:
> I've tested and the issue still isn't fixed on my end...

So i found what was the issue. The fix works when we add an empty review with a grade of 0. On the other hand, a non empty review continue to create the bug. I will look into it.

Pijukai:
Now it should work more consistently but the case where grade is different of 0 and review is not empty string still create the bug

sean-prz (changes requested):
Hey unfortunately I tested multiples cases and each time after adding a review the problem is still here. It is not at all fixed on my end

LouisJamesVasseur (approved):
Your most recent commit fixed the issue! Great job!

sean-prz:
I'll just wait until Maria fixes the problem on the main branch before approving so that we don't have even more stuff to revert

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -81,12 +81,14 @@ fun MapScreen(
   val mapViewportState = MapConfig.createMapViewPortStateFromViewModel(mapViewModel)
   var removeViewAnnotation = remember { true }
   var cancelables = remember<Cancelable> { Cancelable({}) }
+  var listener = remember<MapIdleCallback?> { null }
   var pointAnnotationManager by remember { mutableStateOf<PointAnnotationManager?>(null) }
+  val selectedParking = parkingViewModel.selectedParking.collectAsState().value
```
sean-prz, line 86: I think calling .collectAsState().value is useless, if the state is updated the selectedParking val is not updated when writing this.
you should just use the `by` syntax

sean-prz (commented):
https://github.com/user-attachments/assets/9cbcf1e3-1760-48e8-912e-4462d9f4fa41

It still does not work on my end, sorry :( Maybe see the comment above I don't think your last commit adds anything

sean-prz:
I rebased the branch from main and it seems to work now

sean-prz:
Rebase the remote branch and re-request me when it's done. I'll have a final look

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Pijukai: Well, technically speaking we collect a State and then I extract the `value` attribute. Collecting a state first means that it will be updated on change and then extract the `value` no ?

LouisJamesVasseur: I agree w/ @Pijukai here. This seems like the most sound way to get the selectedParking to update given how the updateParking() function is programmed

sean-prz (approved):
Now it works ! Magic no idea why rebasing fixed it.

Make sure to **squash merge** so that your work get appended on the main, I'm afraid it will be broken if we do a normal merge as the order of commit will be different than what we tested, even if that is not supposed to break it.

Pijukai:
> Now it works ! Magic no idea why rebasing fixed it.
> 
> Make sure to **squash merge** so that your work get appended on the main, I'm afraid it will be broken if we do a normal merge as the order of commit will be different than what we tested, even if that is not supposed to break it.

I rebased like your previous comment suggested to make sure it works for everyone.

PR #138, opened 2024-11-02
Feature toggle button

### LINKED ISSUE #135: Create a toggle button as Atom

andrepeiry:
Add a new Atom : ToggleButton (here with the two states displayed on top of one another)
image

It can be use to display a filter, or anything that has to be toggle on an off.

This PR also solve the Button (base one) not being turned gray when disabled.

image

Close #135

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/theme/atoms/Button.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/theme/atoms/Button.kt
@@ -1,6 +1,7 @@
 package com.github.se.cyrcle.ui.theme.atoms
 
 import androidx.compose.material3.Button
+import androidx.compose.material3.ButtonDefaults
 import androidx.compose.material3.ExtendedFloatingActionButton
 import androidx.compose.material3.FloatingActionButton
 import androidx.compose.material3.Icon
@@ -19,6 +20,7 @@ import com.github.se.cyrcle.ui.theme.disabledColor
 import com.github.se.cyrcle.ui.theme.getButtonColors
 import com.github.se.cyrcle.ui.theme.getColor
 import com.github.se.cyrcle.ui.theme.getOnColor
+import com.github.se.cyrcle.ui.theme.onDisabledColor
 
 /**
  * Create a themed small floating action button, with simplified arguments.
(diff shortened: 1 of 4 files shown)


sean-prz (approved):
I appreciate the effort that has gone into creating a library of atoms.
However, I feel that this approach may be somewhat excessive. 
Perhaps a more effective strategy would be for each member to contribute the atoms they have created for their screens. This way, you could review the contributions to ensure they align with the correct colors, themes, and overall consistency, while also ensuring that the features remain intact (wink)

app/src/main/java/com/github/se/cyrcle/ui/theme/atoms/Button.kt
```diff
@@ -156,7 +158,36 @@ fun Button(
...
+@Composable
+fun ToggleButton(
+    text: String,
+    value: MutableState<Boolean>,
+    modifier: Modifier = Modifier,
+    colorLevel: ColorLevel = ColorLevel.PRIMARY,
+    testTag: String = "ToggleButton"
+) {
+  Button(
+      onClick = { value.value = !value.value },
```
Armetiis, line 184: I think this is too restrictive. Specifically, it is not always the case that this button is going to get called with a mutable state of boolean. As an example, in the case of filter, I'd like it to act on a set. I think it is imperative to have a customazible onclick function

Armetiis (changes requested):
The way this is done is very clean, nothing to say. However, I am unsatisfied with the usability  of this atom. More details in the comment I left

andrepeiry:
> I appreciate the effort that has gone into creating a library of atoms. However, I feel that this approach may be somewhat excessive. Perhaps a more effective strategy would be for each member to contribute the atoms they have created for their screens. This way, you could review the contributions to ensure they align with the correct colors, themes, and overall consistency, while also ensuring that the features remain intact (wink)

At first, this PR was bound to be a fix on Button, but after more discussion, it was discovered that it wasn't the direction  this PR was supposed to take, as instead of having a deactivating button, it was one that could Toggle a value. But I agree that everyone should do his own atom & molecule, even though that would mean making them generic (we don't want to make three time the same element for three different usage).

app/src/main/java/com/github/se/cyrcle/ui/theme/atoms/Button.kt
andrepeiry: The functionality has been added in latest commit ^^

Armetiis (approved):
I will approve but I will make a new PR with two variants. One just like you had in the beggining and another one that takes a boolean rather than a state. It doesn't make much sense to pass a mutable state when the onclick function is custom. In the example i gave you in the previous commit, that would just imply having to create a new mutable state, and adding the logic to update it. I'm sorry about this, I should've been the one creating this atom in the beggining, rather than making you go in details like this

PR #139, opened 2024-11-02
Rework list screen UI and add infinite scroll

### LINKED ISSUE #100: Rework list screen UI and make infinite scroll
### Problem
At this stage, the list screen only loads 5 nearest parkings rather than dynamically loading more and more parkings as we scroll.

- Add fetch functionality when we reach bottom of scroll using lazylist

Also rework ui of the screen

Armetiis:
Closes #100 

### What
This PR does two things:
1. Rework the UI of the list screen, here is a preview
image
image

2. Make the list screen infinite scroll

### Key changes
For consistency while scrolling, the filtering system had to be changed. We are back to a classic filtering system, as the smart filtering system and the infinite scroll are not compatible. (Previously, the parkings were first sorted by number of matched criterion and then by distance. This would mean that while scrolling down, when we would reach the bottom, parkings that were further but had more matched criterion would get propulsed to the top of the list.)

### Coverage 
image

### Tests
Tests have been adapted to cope with the new UI.

### Future work:
Once the StarReview molecule is done, it shall be added in place of the text "Rating:"
Once the logic for the image database is done, we can add an image to preview the parking

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
@@ -39,11 +39,6 @@ class ParkingViewModel(
   private val tilesToParking =
       MutableStateFlow<LinkedHashMap<Tile, List<Parking>>>(LinkedHashMap(10, 1f, true))
 
-  init {
-    // TODO: Replace with actual location
-    parkingRepository.onSignIn { getKClosestParkings(Point.fromLngLat(6.9, 46.69), 5) }
-  }
-
   /**
    * Fetches the image URL from the cloud storage, This function as to be called after retrieving
    * the path from the Firestore database.
@@ -71,12 +66,7 @@ class ParkingViewModel(
    */
   fun addParking(parking: Parking) {
     parkingRepository.addParking(
-        parking,
-        {
-          val tile = Tile.getTileFromPoint(parking.location.center)
-          tilesToParking.value[tile] = tilesToParking.value[tile]?.plus(parking) ?: listOf(parking)
-        },
-        { Log.e("ParkingViewModel", "Error adding parking", it) })
+        parking, {}, { Log.e("ParkingViewModel", "Error adding parking", it) })
   }
 
   /**
@@ -96,8 +86,6 @@ class ParkingViewModel(
    * @param endPos the opposite corner of the rectangle
    */
   fun getParkingsInRect(startPos: Point, endPos: Point) {
-    // flush the list of parkings
-    _rectParkings.value = emptyList()
     if (startPos.latitude() == endPos.latitude() || startPos.longitude() == endPos.longitude()) {
       Log.e("ParkingViewModel", "Invalid rectangle")
       return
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/se/cyrcle/ui/theme/atoms/Button.kt
```diff
@@ -193,6 +192,36 @@ fun ToggleButton(
       }
 }
 
+/**
+ * Create a themed toggle button. This overload offers more flexibility.
+ *
```
andrepeiry, line 197: Missing  `@param text`

app/src/main/java/com/github/se/cyrcle/ui/theme/atoms/Button.kt
```diff
@@ -193,6 +192,36 @@ fun ToggleButton(
       }
 }
```
andrepeiry, line 193: Overload + modify an unused atom seems a bit overkill, maybe it would be simpler to tune the atom to your needs.

Armetiis: Tried to. The two are not really compatible.

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
```diff
@@ -239,83 +243,103 @@ fun FilterSection(
...
-      content()
-    }
-  }
+        if (isExpanded) {
+          content()
+        }
+      }
 }
 
+@SuppressLint("DefaultLocale")
```
andrepeiry, line 265: I'm not sure that explicitly removing a warning is a good thing, moreover when Android Studio is telling you "*Implicitly using the default locale is a common source of bugs: Use `String.format(Locale, ...)` instead*".

app/src/main/java/com/github/se/cyrcle/ui/theme/atoms/Button.kt
Armetiis: I've looked it at multilple angles and this is what makes the most sense.

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
LouisJamesVasseur: Sometimes you kind of have to suppress warnings, but in this case since Android studio explicitly suggests an alternative it might be good to use it just in case

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
```diff
@@ -33,11 +33,6 @@ class ParkingViewModel(
   private val _selectedParking = MutableStateFlow<Parking?>(null)
   val selectedParking: StateFlow<Parking?> = _selectedParking
 
-  // TODO: Replace with actual location
```
LouisJamesVasseur, line 36: Why is this removed? Is this linked to the feature you're implementing in this PR?

app/src/main/java/com/github/se/cyrcle/ui/theme/atoms/Button.kt
LouisJamesVasseur: On the principle I agree w/ @andrepeiry but I think that the way that the code is implemented here works just fine

LouisJamesVasseur:
Great work, the screen and the code both look great!

LouisJamesVasseur (approved):
Great work, the screen and the code both look great!

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
Armetiis: Yes. I added a LaunchedEffect in the ListScreen. There's no reason between creating the viewmodel and loading parkings. The loading of the k closest parkings should be handled by the screen that uses it

LouisJamesVasseur: Makes sense, thanks for the explanation

andrepeiry (approved):
Everything seems to be working, nice work ^^

andrepeiry (approved):

PR #140, opened 2024-11-03
Bugfix : display newly added spot on the map

sean-prz:
_closes #136_
_closes #143_ 
### What
Fixes a bug introduces by the Tile Manager in which a newly added parking spot, was not displayed on the map as the information about the tile on which it was added was already cached.
Fixes another bug where a spot lying at the intersections of multiple tiles is rendered multiples times
### Why
The user should see the spot that he just added as a confirmation feedback without having to reload the app 
### How
The addParking function of the viewmodel has been changed to add the new parking to the tile parking's list.
The second bug is fixed by going with strict bound instead of non strict bounds for the coordinates
-- -
image

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
@@ -73,7 +73,11 @@ class ParkingViewModel(
     val tile = Tile.getTileFromPoint(parking.location.center)
     tilesToParking.value[tile] = tilesToParking.value[tile]?.plus(parking) ?: listOf(parking)
     parkingRepository.addParking(
-        parking, {}, { Log.e("ParkingViewModel", "Error adding parking", it) })
+        parking,
+        {
+          tilesToParking.value[tile] = tilesToParking.value[tile]?.plus(parking) ?: listOf(parking)
+        },
+        { Log.e("ParkingViewModel", "Error adding parking", it) })
   }
 
   /**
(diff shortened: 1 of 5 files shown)


Armetiis (changes requested):
I think there's an important change to make. The parking has to be added to the tilesToParking map ONLY if the call to addParking is succesful. In this version, if the call to addParking fails, the tilesToParking's value for the tile will be incorrect, as it will contain a parking that doesn't exist

sean-prz:
> I think there's an important change to make. The parking has to be added to the tilesToParking map ONLY if the call to addParking is succesful. In this version, if the call to addParking fails, the tilesToParking's value for the tile will be incorrect, as it will contain a parking that doesn't exist.
> 
> Therefore I think we should make the call in the onsuccess callback

Nicely spotted, I will make this pertinent change asap and rerequest your review

Armetiis (approved):
All good, thanks for the fix

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
```diff
@@ -70,8 +70,14 @@ class ParkingViewModel(
    * @param parking the parking to add
    */
   fun addParking(parking: Parking) {
+    val tile = Tile.getTileFromPoint(parking.location.center)
+    tilesToParking.value[tile] = tilesToParking.value[tile]?.plus(parking) ?: listOf(parking)
```
matttsss, line 74: With regards to what Maria said, we still add it to the tiles even on failliure?

sean-prz: Oops forgot to remove the old implementation, thankfully you checked my work

matttsss (approved):
Looks all good with me, thanks for the work!

Issue #109, opened 2024-10-29
Google accounts and Users are not connected


Armetiis:
We need to establish a connection between the Google account used for Firebase Authentication and the corresponding user object in our database. This linkage is crucial for properly setting the current user in the UserViewModel upon login.

### Current Situation
There's no automatic association between the Firebase Auth user and our database user.

### What we want
When a user logs in with their Google account, we should be able to:
- Retrieve the corresponding user object from our database.
- Set this user as the current user in the UserViewModel.

matttsss:
Closed by #142

Issue #128, opened 2024-11-01
Restrain UI for anonymous users


matttsss:
Lock LoginScreen, AddParkingScreen and AddReviewScreen if the user is signed in anonymously.
Sugest him/her to login or create an account.

PR #142, opened 2024-11-03
Feature add firebase rules

### LINKED ISSUE #108: Manage user perms on the database
- Users should be able to load any parking
- They should be able to load any review (when loading from the concerned parking)
- They shouldn't be able to load any user. This causes a problem:
> Reviews have a userId as argument. If we can't load the user, how do we display basic info (username and profile picture)?

matttsss:
### What is this PR?
This PR is the first step in linking users to the database

### How?
This was done by adding new rules to the Firebase Cloud Storage and linking the `User` dataclass with the loged in user via their uid.

Closes #108 #109 
### Firebase rules: 
```
rules_version = '2';

service cloud.firestore {
  match /databases/{database}/documents {
		
    match /parkings/{parking} {
    	allow read: if request.auth != null;
    	allow create, write: if 
      	request.auth != null && 
        request.auth.token.firebase.sign_in_provider != 'anonymous';
    }
    
    match /users/{user} {
    	allow create, read, write: if 
        request.auth != null && 
        request.auth.token.firebase.sign_in_provider != 'anonymous' &&
        request.auth.uid == request.resource.data.userId;
    }
    
    match /reviews/{review} {
    	allow read: if request.auth != null;
      allow write, delete: if 
      	request.auth != null && 
        request.auth.token.firebase.sign_in_provider != 'anonymous' && 
        request.auth.uid == request.resource.data.owner;
      allow create: if 
      	request.auth != null && 
        request.auth.token.firebase.sign_in_provider != 'anonymous' && 
        request.auth.uid == request.resource.data.owner;
    }
  }
}
```

### Test coverage: 
image


app/src/main/java/com/github/se/cyrcle/model/user/UserRepository.kt
```diff
@@ -10,11 +10,11 @@ interface UserRepository {
   fun onSignIn(onSuccess: () -> Unit)
 
   /**
-   * Get a new unique identifier for a parking
+   * Get a a unique identifier for a user
    *
-   * @return a new unique identifier for a parking
+   * @return A UID linked to the user's signed in account
    */
```
sean-prz, line 16: If you think this is a better name you should change it across all repository and viewmodel for consitency, I don't really think which name we use matter, but it should be the same accross the project

app/src/main/java/com/github/se/cyrcle/model/user/UserRepositoryFirestore.kt
Armetiis, line 1: I am unsure wether these changes are ideal or not. Wouldn't it be better to let the firebase auth out of the FirestoreRepository and include it in the viewModel. Therefore the function getUid of the viewmodel would directly interact with the auth ? Or have a repo for the auth?
I might be wrong so feel free to explain your point of view

matttsss: The authentication process is included with Firestore and I feel that for abstraction, it is better to have that in our Firebase Repository

sean-prz (commented):
While testing the feature and before diving into the code, I found that indeed an anonymous user could not add spots or review anymore to the database. 
But there is no feedback ! The anonymous user doesn't know that his review can't be added. Could you add a toast or mention in your PR with an issue if this is planned for a future PR .

sean-prz (commented):
There is also the fact that a google-signin user must relaunch the app to see his review persistently (leaving and coming back to the card the review does not display). 

Unsure if these are the new rules that creates this bug since I can't revert them to test how it was before. Anyways we should create an issue to not forget this bug

matttsss:
@sean-prz yes, for the moment that error is loged and intensional. UI blocking for anonymous users is planned in #128

sean-prz (approved):
Thank you for this feature, it really  brings us to a closer to a complete first version.

I would love more details in the description of what you're planning next, since this is the first of a linked series of PR.

app/src/main/java/com/github/se/cyrcle/model/user/UserRepositoryFirestore.kt
Pijukai: Concerning the fact that the view model must manage data and the repository is only for storing it, i feel the same as Armetiis. Now it is not a blocking point for me but can you just detail a bit about the positives points of having it in the FirestoreRepository ?

Pijukai (approved):
Thank you for this new feature because it is getting us closer to a functional community driven app! I agree with Sean, I am interested in having more details about the content of the future PRs in description. Good Job!

app/src/main/java/com/github/se/cyrcle/model/user/UserRepositoryFirestore.kt
matttsss: Sure, they do store data, however, they also provide services. In our case both are entagled within Firebase. So it makes sense to me to keep it "bundled" together within the repository. Now I do agree that this is not optimal, and I could well imagine us creating an authenticator class to help separate these two sometime in the future.

PR #148, opened 2024-11-04
Restrain Anonymous Users

### LINKED ISSUE #128: Restrain UI for anonymous users

matttsss:
### What is this PR?
This PR aims to restrain the UI possibilities offered to Anonymous users.
This is useful for blocking spam and bots since now users are required to login.

### How?
A `Flow<Boolean>` was added as attribute to the `UserViewModel` to detect when a user is signed in.
The `ProfileViewScreen` was also refactored: it now leads to a mock `CreateProfileScreen` when the user is not signed in.
Later on, this screen will have to be filled to offer the user to create his account and check if it doesn't exist already.

### Test coverage
image

Closes #128


app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -73,17 +75,20 @@ const val LAYER_ID = "0128"
...
+  val enableParkingAddition by userViewModel.isSignedIn.collectAsState(false)
+
   val mapViewportState = MapConfig.createMapViewPortStateFromViewModel(mapViewModel)
   var removeViewAnnotation = remember { true }
-  var cancelables = remember<Cancelable> { Cancelable({}) }
+  var cancelables = remember { Cancelable {} }
   var listener = remember<MapIdleCallback?> { null }
   var pointAnnotationManager by remember { mutableStateOf<PointAnnotationManager?>(null) }
-  val selectedParking = parkingViewModel.selectedParking.collectAsState().value
+  val selectedParking by parkingViewModel.selectedParking.collectAsState()
```
sean-prz, line 91: @LouisJamesVasseur @Pijukai  😉

app/src/main/java/com/github/se/cyrcle/ui/profile/CreateProfileScreen.kt
```diff
@@ -0,0 +1,51 @@
...
+
+  Scaffold(
+      modifier = Modifier.fillMaxSize().testTag("CreateProfileScreen"),
+      topBar = { TopAppBar(navigationActions, screenTitle) }) { padding ->
+        Column(
+            modifier = Modifier.fillMaxSize().padding(padding),
+            verticalArrangement = Arrangement.Center) {
+              Text("TODOOOOOOOOOO")
+
+              // GoogleSignInButton(
```
LouisJamesVasseur, line 30: Any reason why all this is commented? Is this a feature to roll out in a future PR?

LouisJamesVasseur (approved):
Great work! The screens look nice. We'll need to make another PR to implement the SignedOut version of the screen

app/src/main/java/com/github/se/cyrcle/ui/profile/CreateProfileScreen.kt
matttsss: This was just to test but it cannot be un-commented as is, it does not make any checks for a user already existing in the database and does not fill in all the fields for the User. I left it there as an example and help for future work.

LouisJamesVasseur: Good call then (for next time maybe make it a block comment so it's easier to uncomment :)) )

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Pijukai: @sean-prz That's just to simplify but at the end it doesn't change anything 😉

app/src/main/java/com/github/se/cyrcle/ui/map/overlay/ZoomControls.kt
```diff
@@ -18,10 +18,11 @@ import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.unit.dp
 
 @Composable
-fun ZoomControls(onZoomIn: () -> Unit, onZoomOut: () -> Unit) {
+fun ZoomControls(modifier: Modifier = Modifier, onZoomIn: () -> Unit, onZoomOut: () -> Unit) {
```
Pijukai, line 21: I think you can remove the type `Modifier`  here since you are already declaring the value

Pijukai (approved):
Great work thanks for that ! Just don't forget to mention what kind of functionality is restrained to anonymous user by this PR.

Issue #130, opened 2024-11-01
Integrate User location on the Map


Pijukai:
We need to integrate the localisation on the MapScreen. This implies : 

- [x] Handling the user localisation
- [x] Displaying the localisation as a Puck 
- [x] Displaying only when permission was granted

PR #150, opened 2024-11-04
Geolocalisation of the user

### LINKED ISSUE #129: Permission Handling for Location
We need to ask the user for permission, obtain it (if the user is okay) and then handle it: 

- [x]  Set up MapBox for localisation 
- [x]  Ask the user for his localisation
- [x]  Handle the case where the user say yes 
- [x]  Handle the case where the user say no

Pijukai:
Closes #129 #130 

## Content of the PR

This PR implements : 

- Asks the user for his location (Approximate or precise)
- Display the user position if he agreed
- Display EPFL location as default position

## How ? 

To do that I created a class wrapper `Permission Handler` to implement the `PermissionListener` and `PermissionManager` from MapBox that handle this feature. This class is invoked at MainActivity. 

## Why ? 

This app is location-based and it allows the user to see relevant bike parking based on its location. 

## Testing

This feature needs to be tested manually. When testing, be sure to close (make sure it stops and is not running without you knowing) and open the app correctly and be careful of the OS interpreting your choices correctly in the settings. 

A few testing case : 

- If the user did not allow then verify that the app does not ask again for the permission. 
- If the user allows only for this time then make sure the app asks everytime the app closes. 
- If the user allows when using the app then make sure the app does not asks again for permission. 
- Try to change the permissions for localisation from the settings to see how the app adapts

There was a problem with adding geolocalisation during testing because it made some android tests fail. This was fixed by enabling dependency injection for the `PermissionHandler`

## Bug 

There is a known bug which can be reproduced using the following steps : 

Starting app > Granting Permission (always grant) > Sign In > No localisation displayed > Restarting the app > Sign In > Localisation displayed (#159)

This will be fixed during next sprint as well as handling when the user move. (#160)

PS : Emulators are located in Mountain View at Google Headquarters. 

## Code Coverage 
image

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/CyrcleNavHost.kt
+++ b/app/src/main/java/com/github/se/cyrcle/CyrcleNavHost.kt
@@ -11,12 +11,12 @@ import com.github.se.cyrcle.model.user.UserViewModel
 import com.github.se.cyrcle.ui.addParking.attributes.AttributesPicker
 import com.github.se.cyrcle.ui.addParking.location.LocationPicker
 import com.github.se.cyrcle.ui.authentication.SignInScreen
-import com.github.se.cyrcle.ui.card.CardScreen
 import com.github.se.cyrcle.ui.list.SpotListScreen
 import com.github.se.cyrcle.ui.map.MapScreen
 import com.github.se.cyrcle.ui.navigation.NavigationActions
 import com.github.se.cyrcle.ui.navigation.Route
 import com.github.se.cyrcle.ui.navigation.Screen
+import com.github.se.cyrcle.ui.parkingDetails.ParkingDetailsScreen
 import com.github.se.cyrcle.ui.profile.ProfileScreen
 import com.github.se.cyrcle.ui.review.AllReviewsScreen
 import com.github.se.cyrcle.ui.review.ReviewScreen
@@ -45,7 +45,9 @@ fun CyrcleNavHost(
         route = Route.LIST,
     ) {
       composable(Screen.LIST) { SpotListScreen(navigationActions, parkingViewModel) }
-      composable(Screen.CARD) { CardScreen(navigationActions, parkingViewModel, userViewModel) }
+      composable(Screen.CARD) {
+        ParkingDetailsScreen(navigationActions, parkingViewModel, userViewModel)
+      }
       composable(Screen.REVIEW) {
         ReviewScreen(navigationActions, parkingViewModel, reviewViewModel, userViewModel)
       }
--- a/app/src/main/java/com/github/se/cyrcle/CyrcleNavHost.kt
+++ b/app/src/main/java/com/github/se/cyrcle/CyrcleNavHost.kt
@@ -58,7 +58,7 @@ fun CyrcleNavHost(
         startDestination = Screen.MAP,
         route = Route.MAP,
     ) {
-      composable(Screen.MAP) { MapScreen(navigationActions, parkingViewModel, mapViewModel,userViewModel, permissionGranted = permissionGranted) }
+      composable(Screen.MAP) { MapScreen(navigationActions, parkingViewModel,userViewModel, mapViewModel, permissionGranted = permissionGranted) }
     }
 
     navigation(startDestination = Screen.LOCATION_PICKER, route = Route.ADD_SPOTS) {
--- a/app/src/main/java/com/github/se/cyrcle/CyrcleNavHost.kt
+++ b/app/src/main/java/com/github/se/cyrcle/CyrcleNavHost.kt
@@ -58,7 +58,14 @@ fun CyrcleNavHost(
         startDestination = Screen.MAP,
         route = Route.MAP,
     ) {
-      composable(Screen.MAP) { MapScreen(navigationActions, parkingViewModel,userViewModel, mapViewModel, permissionGranted = permissionGranted) }
+      composable(Screen.MAP) {
+        MapScreen(
+            navigationActions,
+            parkingViewModel,
+            userViewModel,
+            mapViewModel,
+            permissionGranted = permissionGranted)
+      }
     }
 
     navigation(startDestination = Screen.LOCATION_PICKER, route = Route.ADD_SPOTS) {
--- a/app/src/main/java/com/github/se/cyrcle/CyrcleNavHost.kt
+++ b/app/src/main/java/com/github/se/cyrcle/CyrcleNavHost.kt
@@ -1,5 +1,3 @@
-package com.github.se.cyrcle
-
 import androidx.compose.runtime.Composable
 import androidx.navigation.NavHostController
 import androidx.navigation.compose.NavHost
@@ -32,6 +30,7 @@ fun CyrcleNavHost(
     userViewModel: UserViewModel,
     mapViewModel: MapViewModel,
     addressViewModel: AddressViewModel,
+    permissionGranted: Boolean
 ) {
   NavHost(navController = navController, startDestination = Route.AUTH) {
     navigation(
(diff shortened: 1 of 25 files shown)


app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -78,7 +83,8 @@ fun MapScreen(
     parkingViewModel: ParkingViewModel,
     userViewModel: UserViewModel,
     mapViewModel: MapViewModel,
-    zoomState: MutableState<Double> = remember { mutableDoubleStateOf(defaultZoom) }
+    zoomState: MutableState<Double> = remember { mutableDoubleStateOf(defaultZoom) },
+    permissionGranted: Boolean
```
sean-prz, line 87: For Backward compatibility with all our old test, you should have put a default value of false here for permissionGranted.

This would have saved you from going modifying all of our actual test that use a map

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -326,3 +339,29 @@ private fun drawMarkers(
...
+ * @param mapViewModel the MapViewModel to update the camera position
+ */
+private fun initLocationComponent(mapView: MapView, mapViewModel: MapViewModel) {
+  val locationComponentPlugin = mapView.location
+  locationComponentPlugin.updateSettings {
+    this.enabled = true
+    this.locationPuck = createDefault2DPuck(true)
+  }
+  val onIndicatorPositionChangedListener = OnIndicatorPositionChangedListener { location ->
+    mapView.mapboxMap.setCamera(CameraOptions.Builder().center(location).build())
```
sean-prz, line 356: This is annoying, each time our position Changes the camera is centered around us. 

We're too close to the deadline to change it. So can you already create an issue and mention this in the PR please

app/src/main/java/com/github/se/cyrcle/MainActivity.kt
```diff
@@ -59,7 +67,8 @@ class MainActivity : ComponentActivity() {
               reviewViewModel,
               userViewModel,
               mapViewModel,
-              addressViewModel)
+              addressViewModel,
+              PermissionsManager.areLocationPermissionsGranted(this))
```
sean-prz, line 71: I believe this is why you need to relaunch the app for the permission to work. This runs as soon as you create the activity aka launch the app. So you haven't had the chances to accept permission. 

The solution should be to delay this line of code

sean-prz (approved):
Thanks for the quantity of comments. I would have been lost otherwise.

Except for the two bugs mentionned, it works on my phone.
I am not gonna request changes for these bugs as you mentioned them in your PR and we are running out of time but please can you already create the issues for the 2 bugs introduced by this PR and link them to the description of this PR. 

And maybe add the default parameter to the map screen when you fix the bugs, it will make testing more convinient

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
sean-prz: Actually this could be a feature, we could have two modes one that follow us and one that doesn't. But like this is just too constraining

app/src/main/java/com/github/se/cyrcle/MainActivity.kt
Pijukai: That's a very insightful comment I will make sure to test your hypothesis next sprint !

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Pijukai: Yes that's what i have planned , add a button to center on the puck and not centering on the puck otherwise. Will be relevant for the user. This is part of the task `handling when user move` for next Sprint.

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Pijukai: Yess i will make sure to change that when fixing the bug thanks

Pijukai:
> Thanks for the quantity of comments. I would have been lost otherwise.
> 
> Except for the two bugs mentionned, it works on my phone. I am not gonna request changes for these bugs as you mentioned them in your PR and we are running out of time but please can you already create the issues for the 2 bugs introduced by this PR and link them to the description of this PR.
> 
> And maybe add the default parameter to the map screen when you fix the bugs, it will make testing more convinient

Will do ! Thank you for the review.

app/src/main/java/com/github/se/cyrcle/MainActivity.kt
```diff
@@ -36,10 +39,15 @@ class MainActivity : ComponentActivity() {
...
   @Inject lateinit var addressRepository: AddressRepository
 
+  @Inject lateinit var permissionsHandler: PermissionHandlerInterface
+
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
     requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_PORTRAIT
 
+    permissionsHandler.initHandler(this)
+    Log.e("MainActivity", "PermissionsHandler initialized")
```
andrepeiry, line 49: I don't think logging this as an error (`.e`) is what is intended here.

app/src/main/java/com/github/se/cyrcle/CyrcleNavHost.kt
```diff
@@ -62,7 +61,12 @@ fun CyrcleNavHost(
         route = Route.MAP,
     ) {
       composable(Screen.MAP) {
-        MapScreen(navigationActions, parkingViewModel, userViewModel, mapViewModel)
+        MapScreen(
+            navigationActions,
+            parkingViewModel,
+            userViewModel,
+            mapViewModel,
+            permissionGranted = permissionGranted)
```
andrepeiry, line 69: [Non-blocking] Why specify only this argument ?

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Pijukai: Done !

app/src/main/java/com/github/se/cyrcle/CyrcleNavHost.kt
Pijukai: because we have the zoom state before in the list of argument which would push me to either add `zoom = defaultzoom` or do that

app/src/main/java/com/github/se/cyrcle/MainActivity.kt
Pijukai: Yep i will remove it thanks

Pijukai: Done !

app/src/main/java/com/github/se/cyrcle/MainActivity.kt
andrepeiry, line 70: I may have multiple clues
- Permission "asker", at least in my emulator, let me choose between *Precise* and *Approximate* location. How does that works with `areLocationPermissionsGranted` ? The name of the function does not help. It seems that the Mapbox documentation isn't helping much either
- I moved `PermissionsManager.areLocationPermissionsGranted(this)` right after the permission handler initialization (`:49`), and printing it returns me `false` during my first try.

Pijukai: MapBox is managing the difference internally while we are left with only the approval or not from the user I think. 
For the second point, yes sean mentionned that it could come from the fact that it is compiled at the same time not taking into account the user choice until restart.

app/src/main/java/com/github/se/cyrcle/CyrcleNavHost.kt
andrepeiry: Missed that, sorry for the comment

andrepeiry (approved):
I think that the short-term solution for the bugfix is acceptable, but it will need to be a P1 next sprint, as M2 is at the end. Rest of the code seems to be in order to my eyes. Good job !

Pijukai:
> I think that the short-term solution for the bugfix is acceptable, but it will need to be a P1 next sprint, as M2 is at the end. Rest of the code seems to be in order to my eyes. Good job !

I will do this as a P1 task next sprint no worries.

PR #151, opened 2024-11-04
Integrate user viewmodel in profile screen

### LINKED ISSUE #155: Implement Viewmodels to user profile screen

AlexandreBeeDeeBop:
Content of the PR
This PR integrates the user viewmodel in the profile screen.

How ?
By using the functions already created in UserViewModel.

Why ?
Integrating the UserViewModel allows for centralized management of user-related data, improving code maintainability and readability and it provides a reactive data flow, ensuring the UI updates automatically when user data changes.

Testing
There are a lot of tests, for literally every action you can possibly think of. Some of them are just remakes of what was done back when the profile screen consisted of hardcoded values instead of a mock viewmodel.

Code Coverage
image

Closes #155


matttsss:
Please don't forget to maintain your branch up to date, you are missing commits from maint that you might have to rebase upon...

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
```diff
@@ -158,36 +143,32 @@ private fun EditProfileContent(
   InputText(
       value = firstName,
       onValueChange = onFirstNameChange,
-      label = stringResource(R.string.view_profile_screen_first_name_label),
+      label = "First Name",
```
matttsss, line 146: Please try and keep constants strings in the `.xml` files like @andrepeiry did before on this screen

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
```diff
@@ -288,20 +267,21 @@ private fun ProfileImage(
 }
 
 @Composable
-private fun FavoriteParkingsSection(
-    favoriteParkings: List<Parking>,
-    onFavoritesUpdated: (List<Parking>) -> Unit
-) {
+private fun FavoriteParkingsSection(userViewModel: UserViewModel) {
+  val favoriteParkings = userViewModel.favoriteParkings.collectAsState().value
```
matttsss, line 271: This line could be simplified by: 
`val favoriteParkings by userviewModel.favoriteParkings.collectAsState()`
Although it might require an import

matttsss (commented):
An overall good job, although there are few comments that you can adress.
Also there is a bug in UI:
After modifying the username and saving, when you go back again into edit mode and then cancel, the username goes back to the initial value and not the new one I just changed and save before.

matttsss:
Also, it would be nice to have a screenshot of the code coverage (it does not show up in the PR description).
And it looks like you didn't link the right issue, number 420 does not exist

app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ProfileScreenMock.kt
```diff
@@ -0,0 +1,303 @@
...
+    // Fetch favorite parkings
+    userViewModel.getSelectedUserFavoriteParking()
+
+    composeTestRule.setContent {
+      ViewProfileScreen(navigationActions = mockNavigationActions, userViewModel = userViewModel)
+    }
+  }
+
+  @Test
+  fun testUserSetupAndFavoriteParkings() {
```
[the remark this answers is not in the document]
AlexandreBeeDeeBop: Because something was wrong with the user view model, which I then fixed thanks to this test. But you're right, I'll still remove it because now everything's good and it's irrelevant.

app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ProfileScreenMock.kt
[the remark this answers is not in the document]
AlexandreBeeDeeBop: You're right, I named it like that before learning about the existence of ViewProfileScreen. Ill change that.

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
AlexandreBeeDeeBop: no it cant ?

AlexandreBeeDeeBop: or do you expect me to call favoriteParkings.value everywhere ? I dont know why i would do that though...

matttsss: The `by` key-word "extracts" the value directly, it's the delegation concept of Kotlin. The value of the variable is delegated to the state.
But it's a minor detail nevermind😅

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
AlexandreBeeDeeBop: Noted

AlexandreBeeDeeBop:
> An overall good job, although there are few comments that you can adress. Also there is a bug in UI: After modifying the username and saving, when you go back again into edit mode and then cancel, the username goes back to the initial value and not the new one I just changed and save before.

Good catch ! The updateUser function only changes the user info in the database but doesnt fetch it to update the state of the currentUser, which is why it wouldnt have worked as I wanted. So for the moment I just call again the getUserById method which updates the currentUser. Do you think I should just call setCurrentUser instead ?

app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ViewProfileScreenTest.kt
```diff
@@ -13,203 +25,289 @@ import org.mockito.Mockito.mock
...
+    composeTestRule.onNodeWithText("Rue de la paix").assertIsDisplayed()
   }
 
   @Test
-  fun testRemoveFavoriteParking() {
-    // Wait for favorite parkings to be displayed
-    composeTestRule.waitUntil(timeoutMillis = 5000) {
-      composeTestRule.onAllNodesWithTag("FavoriteToggle_0").fetchSemanticsNodes().isNotEmpty()
-    }
+  fun testRemoveAllFavoriteParkings() {
```
andrepeiry, line 221: For this test, you're always removing the first parking (because upon removing, the index get modify). Could you also test the case where you remove a parking in the middle as well as the last parking, and check for the indexes ?

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
```diff
@@ -253,7 +240,8 @@ private fun ProfileImage(
               contentAlignment = Alignment.Center) {
                 Icon(
                     imageVector = Icons.Filled.Person,
-                    contentDescription = "Default Profile Picture",
+                    contentDescription =
+                        stringResource(R.string.view_profile_screen_default_profile_picture),
```
andrepeiry, line 244: `contentDescription` are not displayed Strings, so no need to use the `strings.xml` offload

This comment is also valid for the line `256`, `270`, `332`

This comment is not blocking

matttsss (approved):
Alright, the bug for UI is still there (it's the firebase rules that prevent from updating the users in the database). The tests pass since they use the mock repositories that do not have rules.
Since this is not your fault and I have no more issues, this PR is good to go on my side.

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
```diff
@@ -355,8 +342,9 @@ private fun FavoriteParkingCard(parking: Parking, index: Int, onRemove: () -> Un
         title = { Text(stringResource(R.string.view_profile_screen_remove_favorite_dialog_title)) },
         text = {
           Text(
-              stringResource(R.string.view_profile_screen_remove_favorite_dialog_message)
-                  .format(parking.optName))
+              stringResource(
+                  R.string.view_profile_screen_remove_favorite_dialog_message,
+                  parking.optName ?: ""))
```
andrepeiry, line 347: There is a `default_parking_name` String that exists in `strings.xml` that you can use instead of an empty string for an unnamed parking.

andrepeiry (changes requested):
Good implementation of the *UI layer* of the MVVM design pattern for the user. Please fix or discuss the blocking (all that are not tagged as "*Non-blocking*" inline comment ^^

app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ViewProfileScreenTest.kt
AlexandreBeeDeeBop: I added a test where i remove the middle parking, assert the 1st and 3rd (now 2nd) are still here. And then I remove the 3rd (now 2nd) and assert the 1st is still displayed. Is that what you wanted ?

app/src/androidTest/java/com/github/se/cyrcle/ui/card/CardScreenTest.kt
```diff
@@ -95,20 +92,11 @@ class CardScreenTest {
...
-        .onNodeWithTag("RackTypeColumn")
-        .onChildAt(1)
-        .assertTextContains(ParkingRackType.TWO_TIER.description)
-    composeTestRule
-        .onNodeWithTag("ProtectionColumn")
-        .onChildAt(1)
-        .assertTextContains(ParkingProtection.COVERED.description)
+    composeTestRule.onNodeWithTag("CapacityColumn").onChildAt(1).assertTextContains("51-100 spots")
+    composeTestRule.onNodeWithTag("RackTypeColumn").onChildAt(1).assertTextContains("Two-tier rack")
+    composeTestRule.onNodeWithTag("ProtectionColumn").onChildAt(1).assertTextContains("Covered")
```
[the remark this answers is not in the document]
AlexandreBeeDeeBop: I swear i dont remember changing that, Ill blame Louis because it's his code and hope he doesnt see this comment ! 
Done.

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
```diff
@@ -344,7 +344,7 @@ private fun FavoriteParkingCard(parking: Parking, index: Int, onRemove: () -> Un
           Text(
               stringResource(
                   R.string.view_profile_screen_remove_favorite_dialog_message,
-                  parking.optName ?: ""))
+                  parking.optName ?: R.string.default_parking_name))
```
andrepeiry, line 347: You forget to call `stringResource` around `R.string.default_parking_name`, which is only a index (`Int`) that Koltin is probably automatically casting, which might become problematic.

AlexandreBeeDeeBop: Oops, im too distracted. Glad you're here

andrepeiry (approved):
Everything is in order for me now, good job !

PR #158, opened 2024-11-07
Implement Logging for Reviews

LouisJamesVasseur:
### What is this PR?

In this PR, logging is added to Reviews: each review is now connected to a User and a date, allowing for a better view of all the Reviews. Users will now only be able to submit 1 review per parking, which they can then edit by re-clicking on the Parking they added a review to

### Why? 

As Reviews become increasingly present in the app, we need a way to track the who/when of a review. Users now being able to edit their reviews also adds great functionality.

### How?

From the All Reviews Screen, users can now see if they've already added a Review for the Parking (visible on the FAB in the image below):

Screenshot 2024-11-10 at 17 22 51
Screenshot 2024-11-10 at 17 23 16

Text on both screens display "Edit" if User has already added a Review or "Add" if he hasn't.

### Testing

The added functionality has already been tested as a part of other PRs so no new tests are provided now

### Coverage Report

Screenshot 2024-11-10 at 17 21 14

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
@@ -146,12 +146,21 @@ class ParkingViewModel(
    * @param newScore: score of the new review to add
    * @param parking: Parking to update (selectedParking by default, should never be null)
    */
-  fun updateReviewScore(newScore: Double, parking: Parking = selectedParking.value!!) {
-    parking.avgScore =
-        (100 * ((parking.avgScore * parking.nbReviews) + newScore) / (parking.nbReviews + 1))
-            .toInt() / 100.00
-    parking.nbReviews += 1
-    parkingRepository.updateParking(parking, {}, {})
+  fun updateReviewScore(
+      newScore: Double,
+      oldScore: Double = 0.0,
+      parking: Parking = selectedParking.value!!,
+      isNewReview: Boolean
+  ) {
+    if (isNewReview) {
+      parking.avgScore =
+          (100 * ((parking.avgScore * parking.nbReviews) + newScore) / (parking.nbReviews + 1))
+              .toInt() / 100.00
+      parking.nbReviews += 1
+      parkingRepository.updateParking(parking, {}, {})
+    } else {
+      parking.avgScore += (oldScore - newScore) / parking.nbReviews
+    }
   }
 
   // create factory (imported from bootcamp)
--- a/app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
@@ -159,7 +159,8 @@ class ParkingViewModel(
       parking.nbReviews += 1
       parkingRepository.updateParking(parking, {}, {})
     } else {
-      parking.avgScore += (oldScore - newScore) / parking.nbReviews
+      val delta = if (parking.nbReviews != 0) (oldScore - newScore) / parking.nbReviews else 0.0
+      parking.avgScore += delta
     }
   }
 
--- a/app/src/main/java/com/github/se/cyrcle/model/review/Review.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/review/Review.kt
@@ -1,9 +1,12 @@
 package com.github.se.cyrcle.model.review
 
+import com.google.firebase.Timestamp
+
 data class Review(
     val uid: String,
     val owner: String,
     val text: String,
     val rating: Double,
-    val parking: String
+    val parking: String,
+    val time: Timestamp = Timestamp.now()
 )
--- a/app/src/main/java/com/github/se/cyrcle/ui/review/AllReviewsScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/review/AllReviewsScreen.kt
@@ -14,6 +14,7 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.material3.Card
 import androidx.compose.material3.CardDefaults
+import androidx.compose.material3.FloatingActionButton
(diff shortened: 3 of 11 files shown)


app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
```diff
@@ -146,12 +146,21 @@ class ParkingViewModel(
...
+      isNewReview: Boolean
+  ) {
+    if (isNewReview) {
+      parking.avgScore =
+          (100 * ((parking.avgScore * parking.nbReviews) + newScore) / (parking.nbReviews + 1))
+              .toInt() / 100.00
+      parking.nbReviews += 1
+      parkingRepository.updateParking(parking, {}, {})
+    } else {
+      parking.avgScore += (oldScore - newScore) / parking.nbReviews
```
AlexandreBeeDeeBop, line 162: I already did a similar comment on a previous PR of yours, I know it shouldnt happen but maybe create an edge case for parking.nbReviews = 0 just in case.

app/src/main/java/com/github/se/cyrcle/ui/review/AllReviewsScreen.kt
```diff
@@ -31,23 +32,38 @@ import androidx.compose.ui.unit.dp
...
+import com.github.se.cyrcle.model.user.UserViewModel
 import com.github.se.cyrcle.ui.navigation.NavigationActions
+import com.github.se.cyrcle.ui.navigation.Screen
 import com.github.se.cyrcle.ui.theme.molecules.TopAppBar
+import com.google.firebase.Timestamp
+import java.text.SimpleDateFormat
+import java.util.Locale
+
+fun Timestamp.toFormattedDate(): String {
+  val dateFormat = SimpleDateFormat("MMM dd, yyyy", Locale.getDefault())
```
AlexandreBeeDeeBop, line 44: Typo. Also Timestamp.now() is doodoo i remember using it and it's pretty bad, small advice, dont rely on it, test it thoroughly !

app/src/main/java/com/github/se/cyrcle/ui/review/AllReviewsScreen.kt
```diff
@@ -86,6 +100,21 @@ fun AllReviewsScreen(
                                           .testTag("ExpandedReviewBox$index")
                                           .clickable { setSelectedCardIndex(-1) }) {
                                     Column {
+                                      userViewModel.getUserById(curReview.owner)
+                                      Text(
+                                          text =
+                                              "Owner: ${userViewModel.currentUser.value?.username!!}",
```
AlexandreBeeDeeBop, line 106: Can you really pretend "!!" ? Who knows if the user deletes his account... then it could become null.
Instead I think you could use "?: [DELETED_ACCOUNT]" or something like that.

AlexandreBeeDeeBop: (Assuming the only reason a username is null is if the user was deleted)

app/src/main/java/com/github/se/cyrcle/ui/review/AllReviewsScreen.kt
```diff
@@ -124,6 +152,22 @@ fun AllReviewsScreen(
                                             Modifier.fillMaxWidth()
                                                 .padding(16.dp)
                                                 .testTag("ReviewCardContent$index")) {
+                                          Text(
+                                              text =
+                                                  "Owner: ${userViewModel.currentUser.value?.username!!}",
```
AlexandreBeeDeeBop, line 157: same here

app/src/main/java/com/github/se/cyrcle/ui/review/ReviewScreen.kt
```diff
@@ -67,12 +83,12 @@ fun ReviewScreen(
               Text(
                   text =
                       when (sliderValue) {
-                        in 0f..1f -> stringResource(R.string.review_screen_terrible_review)
-                        in 1f..2f -> stringResource(R.string.review_screen_poor_review)
-                        in 2f..3f -> stringResource(R.string.review_screen_bad_review)
-                        3f -> stringResource(R.string.review_screen_average_review)
-                        in 3.5f..4f -> stringResource(R.string.review_screen_good_review)
-                        in 4.5f..5f -> stringResource(R.string.review_screen_great_review)
+                        in 0f..1f -> "Terrible experience!"
```
AlexandreBeeDeeBop, line 86: Why did you change from stringResource to string ?

AlexandreBeeDeeBop (commented):
Overall, the "!!" are too risky for me to accept them. Using double bang operators is dangerous as it can cause NullPointerExceptions. It's better to provide default values / handle nulls gracefully.
Make sure you fix those style issues too, to please André who worked hard on them ;)

AlexandreBeeDeeBop:
I might add that you writing "Users will now only be able to submit 1 review per parking" followed by a screenshot of "Owner: Louis V" having two reviews of a same parking spot is quite the ultimate ragebait.😂😂

AlexandreBeeDeeBop (approved):
Thanks for the changes

Pijukai:
When you logIn as a guest you can add a review using your feature which go against what @matttsss did in his PR by updating the rule to prevent that.

Pijukai:
When I click on see all reviews sometime the app just crashes i don't know if you have the same issue ?

app/src/main/java/com/github/se/cyrcle/model/review/Review.kt
```diff
@@ -1,9 +1,12 @@
...
+import com.google.firebase.Timestamp
+
 data class Review(
     val uid: String,
     val owner: String,
     val text: String,
     val rating: Double,
-    val parking: String
+    val parking: String,
+    val time: Timestamp = Timestamp.now()
```
Pijukai, line 11: We agree that the user is not capable of choosing the time himself ?

Pijukai (approved):
Thanks for all the changes you have made. Seems like all the weird behaviors are gone good Job !

Issue #167, opened 2024-11-08
list screen never looks further than 100m if no parking found in the first 100m.


sean-prz:
#### Problem

On the list screen, if no parking is found on the first call to the repository ( 100m around ), the code breaks.
Nothing is shown to the user. 

#### Solution
If no parkings is found in the first 100m, then ask to search in a larger radius. 
Repeat this reasoning until you reach a maximum radius of 1km at which point you give up.

PR #170, opened 2024-11-10
fix: solve issue 167

sean-prz:
_Closes #167_

## Making the filter settings permanent 
By moving the states related to the filter selection in the viewmodel, the filter is now permanent.
That means if the user leaves and come back to the list Screen the filter is still in place

## Refactoring the way the parkings are fetched in ListScreen.
To address this bug and ensure the list screen is working in all cases, the interaction between the ListScreen (UI) and the Viewmodel has been re-thinked.
 
### Making `getParkingsInRadius` private 
It is not the role of the UI anymore to call `getParkingsInRadius`, instead the UI can performs two actions to change the list of parking it receives from the Viewmodel :
1. It can set the circle center with `setCircleCenter`. This has to be called when the screen is composed
2. It can request more parking when the user scrolled to the end with `incrementRadius`.

The role of increasing the radius if no parkings has been found has shifted from the UI to the viewmodel
-- -

### Changing the viewmodel internally
How the viewmodel gets the parking for the state `closestParkings` has been modified to resolve bug #167. 
This refactor also prevents spam requests from the UI to increase the radius multiple times without having the result of the first call. (Wanting to go from a radius 100m to 300m without having the result of radius 200m) 

#### How
- The hard to read `init {}` block has been removed and instead its functionality has been moved inside `updateClosestParkings`
- An argument `nbRequestLeft` has been added to the viewmodel to keep track of when there are currently request still loading
- `nbRequestLeft`. is used in `updateClosestParkings` to avoid useless computation. 

## Global summary of the process 
This graph shows the process of getting parkings, handled by the viewmodel, when the UI request more Parkings.
image

This graph shows the same process when the user loads the listScreen.
image

-- -
### Displaying on the ListScreen the current radius 
Finally this PR adds a UI element to inform the user the radius of search. 
image
 
-- -

-- -
#### Code coverage
image

-- -
Known issue : 
- Switching to the list screen instantly when launching the app, increase the radius to its maximum.

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
@@ -12,7 +12,10 @@ import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.launch
 
-const val KMTOMETERS = 1000.0
+const val KM_TO_METERS = 1000.0
+const val DEFAULT_RADIUS = 100.0
+const val MAX_RADIUS = 1000.0
+const val RADIUS_INCREMENT = 100.0
 /**
  * ViewModel for the Parking feature.
  *
@@ -35,8 +38,18 @@ class ParkingViewModel(
   private val _selectedParking = MutableStateFlow<Parking?>(null)
   val selectedParking: StateFlow<Parking?> = _selectedParking
 
-  // Circle center and radius for the circle search for the list screen
-  private val _radius = MutableStateFlow(0.0)
+  /**
+   * Radius of the circle around the center to display parkings With a public getter to display the
+   * radius in the TopBar of the ListScreen The default value is DEFAULT_RADIUS stored in meters
+   */
+  private val _radius = MutableStateFlow(DEFAULT_RADIUS)
+  val radius: StateFlow<Double> = _radius
+
+  /**
+   * Center of the circle, normally corresponding to the user's location. When this is null, the
+   * filtering and sorting of parkings is not computing. Hence, we should make sure that this is
+   * null when in the map screen.
+   */
   private val _circleCenter = MutableStateFlow<Point?>(null)
   // List of tiles to display
   private var tilesToDisplay: Set<Tile> = emptySet()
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
```diff
@@ -145,21 +133,31 @@ fun FilterHeader(
     selectedCapacities: Set<ParkingCapacity>,
     onAttributeSelected: (ParkingAttribute) -> Unit,
     onlyWithCCTV: Boolean,
-    onCCTVCheckedChange: (Boolean) -> Unit
+    onCCTVCheckedChange: (Boolean) -> Unit,
+    parkingViewModel: ParkingViewModel = viewModel(factory = ParkingViewModel.Factory)
```
matttsss, line 137: The code seems to work for the moment but please be carefull when passing a default ViewModel, I think it instanciates a new one and thus have different state than the one made in the Main. It seems ok since you don't actualy use the default value.

matttsss (commented):
An overal good improvement! I am justa little perplex at the idea of using the `_rectParkings` list state for both `ListScreen` and `MapScreen`, this could cause performance issues when navigating back and forth between both since the position on the map does not have to be near the users' position, what do you think?

sean-prz:
> An overal good improvement! I am justa little perplex at the idea of using the `_rectParkings` list state for both `ListScreen` and `MapScreen`, this could cause performance issues when navigating back and forth between both since the position on the map does not have to be near the users' position, what do you think?

Of course it could be optimized to not recompute the `_rectParkings` when switching between the screens by making `getParkingsInRect` write into two different state. 

Currently this is really not a bottleneck here as when the user moves the camera around, `_rectParkings` is recomputed  each pixel (hundreds of time per scroll movement) anyways. Recomputing it when switching screen is nothing next to that.
And thanks to the cache it is really fast to recompute it

matttsss (approved):

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
```diff
@@ -189,6 +234,38 @@ class ParkingViewModel(
...
+            .filter { parking ->
+              (_selectedProtection.value.isEmpty() ||
+                  _selectedProtection.value.contains(parking.protection)) &&
+                  (selectedRackTypes.value.isEmpty() ||
+                      _selectedRackTypes.value.contains(parking.rackType)) &&
+                  (_selectedCapacities.value.isEmpty() ||
+                      _selectedCapacities.value.contains(parking.capacity)) &&
+                  (!_onlyWithCCTV.value || parking.hasSecurity)
+            }
+    if (_closestParkings.value.size < 5 || _radius.value == MAX_RADIUS) {
```
Armetiis, line 264: 5 here seems a bit arbitrary. Why 5?

Armetiis (approved):
Making the fitlers permanent is a really great change! You solved the bug nicely, making the whole screen's backend much more efficient. Great job

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
sean-prz: You're right I should use Magic number. 
If the UI is not entierly filled with items,  the code to fetch new one when the user arrives at the end of the list  is never read. I've put 5 randomly, but the correct number should be equal to screenheight/itemheight

Armetiis (approved):

PR #171, opened 2024-11-10
Refactor authentication

matttsss:
### What is this PR?
This PR aims to modularise our authentication process by a abstracting the authentication code into it's own "repository"
This will enable later the End2End tests to test the addition of parkings and reviews as a loged in user via dependency injection.

### How?
For this a "Composable repository" `Authenticator` was created, it exposes composable buttons that provide login functionality.
Those buttons have no interaction with the other repositories and only return a `User` as a potential one, no checks made on it's existence in our database. (This will need to be done as a later feature)

### Test coverage:
image


app/src/main/java/com/github/se/cyrcle/ui/authentication/Authenticator.kt
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * Default composable button for authentication
+     *
+     * @param onClick callback for when the button is clicked
+     */
+    @Composable
+    fun DefaultAuthenticateButton(onClick: () -> Unit) {
+      Button(
+          onClick = onClick,
+          colors = ButtonDefaults.buttonColors(containerColor = Color.White),
```
sean-prz, line 63: Careful with using hard-coded colors value.

sean-prz (approved):
Beautiful work. 
I can't find anything to complain about..

app/src/main/java/com/github/se/cyrcle/ui/authentication/Authenticator.kt
matttsss: It's the template button used in all apps, it should be ok 👌

Armetiis (approved):
Very beautifully done. A `@Composable` repository is a very nice solution to the problem, props on the creativity!

PR #173, opened 2024-11-10
BugFix : Display user localisation 

Pijukai:
## Content of the PR

This PR addresses the bug mentionned in the issue where restarting the app was necessary to see user localisation on the map after he agreed on giving it. 

NB:  I didn't put the the default value to `permissionGranted` argument of `MapScreen` since it was remove because now i have the `PermissionHandler` as argument 

## How ? 

The permission handler was added as an argument of MapScreen. Since it is impossible for the user to access the map without giving consent for localisation first, the permission manager will have time to update.

## Cause 

MapBox is running on a different thread than the UI thread so the Navigation had the time to initialize without waiting for user approval. The method `PermissionManager.isLocationGranted()` in the Main return false by default which is the value used to initialize the map (and `CyrcleNavHost`). That is why when restarting the app, the good value was used to initialize the map. 

## Code Coverage 

image


app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -97,6 +101,8 @@ fun MapScreen(
   var pointAnnotationManager by remember { mutableStateOf<PointAnnotationManager?>(null) }
   val selectedParking by parkingViewModel.selectedParking.collectAsState()
 
+  val context = LocalContext.current as Activity
```
matttsss, line 104: I do not realy understand why you pass down `activity` through the whole app when you can just have it with `LocalContext.current`, does this not work?

Pijukai: It's a mistake on my part, it is not used anywhere so forgot to remove it

matttsss: No, what I meant is can you not use that `context` variable as initializer to the `PermissionHandler`?
Instead of adding an extra argument everywhere

LouisJamesVasseur (commented):
First off, from my manual tests, the problem is solved so congrats! Before I approve, I'd like to make sure I understand exactly what's being done here, the part where the actual permissions are being checked is "moved down" to the MapScreen to ensure that everything is handled in the order you're intending then too. Am I correct?

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Pijukai: I need to get the Context from the Main which is not the same as the one in the Map. Since the permissions are asked in the context of the Main I need to bring it to Map the same way i bring the permission Manager so that i can correctly check if location permissions were given.

Pijukai:
> First off, from my manual tests, the problem is solved so congrats! Before I approve, I'd like to make sure I understand exactly what's being done here, the part where the actual permissions are being checked is "moved down" to the MapScreen to ensure that everything is handled in the order you're intending then too. Am I correct?

Yessexactly. This way the permission manager has time to update and it allows us to bypass the problem of concurrent threads in the Main. Tell me if you need further details

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Pijukai: Ok, i tried and it seemed you were right so i cleaned the code a little, feel free to test it again.

LouisJamesVasseur: I agree w/ @matttsss that removing the extra variables make the code cleaner. I've re-tested to ensure that everything still works and it does!

matttsss (approved):
Great, it's a nice cleanup, thank you for the changes!

LouisJamesVasseur (approved):
Great work, and thank you for the quick fix!

Pijukai:
I reintroduced the activity argument because it made the CI fails otherwise despite the fact that it worked locally. But thanks to you @matttsss i cleaned up quite a bit overall.

matttsss:
It was worth a try, weird though that the tests would run localy

PR #176, opened 2024-11-10
Feature log out

### LINKED ISSUE #163: Add logout button
Enable users to logout

matttsss:
## What is this PR?
This PR aims to offer the possibility to users to sign out from the application
image

## How
A new Button was added to the `Authenticator`, it allows users to sign out

## Code Coverage:
image

Closes #163

### DIFF
--- a/app/src/androidTest/java/com/github/se/cyrcle/ui/authentication/AuthenticatorDefaultTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/ui/authentication/AuthenticatorDefaultTest.kt
@@ -40,4 +40,17 @@ class AuthenticatorDefaultTest {
         .performClick()
     assert(clicked)
   }
+
+  @Test
+  fun testDefaultSignOutBtn() {
+    var clicked = false
+    composeTestRule.setContent { Authenticator.DefaultSignOutButton { clicked = true } }
+
+    composeTestRule
+        .onNodeWithTag("SignOutButton")
+        .assertIsDisplayed()
+        .assertHasClickAction()
+        .performClick()
+    assert(clicked)
+  }
 }
--- a/app/src/main/java/com/github/se/cyrcle/ui/authentication/Authenticator.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/authentication/Authenticator.kt
@@ -47,16 +47,18 @@ interface Authenticator {
   /**
    * Composable button that signs the user in anonymously
    *
+   * @param modifier modifier for the button
    * @param onComplete callback for when sign in is complete
    */
-  @Composable fun SignInAnonymouslyButton(onComplete: () -> Unit)
+  @Composable fun SignInAnonymouslyButton(modifier: Modifier, onComplete: () -> Unit)
 
   /**
    * Composable button that signs the user out
    *
+   * @param modifier modifier for the button
    * @param onComplete callback for when the button is clicked
    */
-  @Composable fun SignOutButton(onComplete: () -> Unit)
+  @Composable fun SignOutButton(modifier: Modifier, onComplete: () -> Unit)
 
   /** Static methods for default composable buttons */
   companion object {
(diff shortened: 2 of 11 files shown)


app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
```diff
@@ -71,53 +72,56 @@ fun ViewProfileScreen(
...
-                    lastName = lastName,
-                    username = username,
-                    profilePictureUrl = profilePictureUrl,
-                    onEditClick = { isEditing = true })
-
-                Spacer(modifier = Modifier.height(24.dp))
-
-                FavoriteParkingsSection(userViewModel)
+        Box(Modifier.fillMaxSize().padding(innerPadding)) {
+          authenticator.SignOutButton {
```
sean-prz, line 76: Can you add some padding please, the button is almost touching my phone's edges.

app/src/main/java/com/github/se/cyrcle/ui/authentication/Authenticator.kt
```diff
@@ -101,8 +111,13 @@ interface Authenticator {
...
               Modifier.padding(16.dp)
                   .border(BorderStroke(1.dp, Color.LightGray), RoundedCornerShape(50))
-                  .height(48.dp)
-                  .testTag("AnonymousLoginButton"))
+                  .height(48.dp),
+          testTag = "AnonymousLoginButton")
+    }
+
+    @Composable
+    fun DefaultSignOutButton(onComplete: () -> Unit) {
```
andrepeiry, line 119: Missing documentation

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
```diff
@@ -71,53 +72,56 @@ fun ViewProfileScreen(
...
+            userViewModel.setCurrentUser(null)
+            navigationActions.navigateTo(Route.AUTH)
```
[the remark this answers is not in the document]
andrepeiry: The solution for this is to call `navigateTo(TopLevelDestination)` instead of `navigateTo(String)`, because the one with `TopLevelDestination` pop the stack of the navigation actions.

app/src/androidTest/java/com/github/se/cyrcle/ui/authentication/AuthenticatorDefaultTest.kt
andrepeiry, line 1: I think a nice test to add (might be more towards end-to-end than unit one, but I feel it would have its place) would be to sign in anonymously and to check for an expected behavior in the profile screen later on (should we be able to Sign out if the user is anonymous ? Is the button disabled ?)

andrepeiry (changes requested):
Would it be thinkable to turn the authentication system into a MVVM pattern ? I feel like the following could be doable :
```mermaid
flowchart LR
    subgraph V ["**View**"]
        sib(Sign In Button)
        sob(Sign Out button)
    end

    subgraph M ["**Model**"]
        c(currentAccount)
        si(Sign In)
        so(Sign Out)
    end
si <--> sib
so <--> sob
```
The approach would be more subtle as the `View` part would not be a screen, but rather only a pair of button.

Also, I'm not a big fan of the term `Authenticator` for the system, as it now go both ways (to me, authenticator would only mean signing in). Maybe something like `AccountManager` would better fit what the class.

To approve this PR, I would like the small inline comment solved, as well as the two themes discussed in this comment.

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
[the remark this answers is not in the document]
matttsss: Yes, that's a good idea, thanks for the solution @sean-prz I think it should fit well in your PR on navigation bugfix!

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
matttsss: Done!

app/src/androidTest/java/com/github/se/cyrcle/ui/authentication/AuthenticatorDefaultTest.kt
matttsss: Yes, that is a good idea, since this concept is susceptible to change quite soon (after our conversation with coaches on Friday), I will implement that during the refactor.

matttsss:
@andrepeiry That is a good idea, the reason for this "Composable repository" is that the authentication code needs to send an `Intent` that is why I did not feel comfortable putting `Composable` code into a repository used by a ViewModel (especially since I think there are some synchronization issues between threads, I think). I will check this possibility with coaches next week and consider a refactor to make this cleaner.

sean-prz (approved):

andrepeiry (approved):
Thank you for the clarifications, everything is in order for me. Good job !

PR #178, opened 2024-11-11
Divide the user information into a public and a private part 

sean-prz:
_closes #181_
In preparation for #180 
### What
This PR refactors the user data class and the way user's informations are stored in the database.

### Why
Security rules in firestore only work on a document-level basis, access to specifc fields can't be restricted.
The only solution is to create a sub-collection of the user document with its private fields.

### How
The `User` data class is now just 2 fields : `UserPublic` and `UserDetails`, which are two data classs each containing all public/private field.

The repository, viewmodel and all code related to user's informations has been adapted for this changes.
The `UserDetails` class of a `user` instances is set to null when reading information of another user, this is handled by the repository. 

#### Other minor changes
- The user is added to the database on logged-in as opposed to when modifying his profile.
- The user can see his private details on the profile screen with peristence.
-- -
### [Updated Firestore rules](https://console.firebase.google.com/u/2/project/cyrcle-ab198/firestore/databases/-default-/rules)

-- -

image

### DIFF
--- a/app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ViewProfileScreenTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ViewProfileScreenTest.kt
@@ -15,6 +15,8 @@ import com.github.se.cyrcle.model.parking.ParkingRackType
 import com.github.se.cyrcle.model.parking.ParkingViewModel
 import com.github.se.cyrcle.model.parking.TestInstancesParking
 import com.github.se.cyrcle.model.user.User
+import com.github.se.cyrcle.model.user.UserDetails
+import com.github.se.cyrcle.model.user.UserPublic
 import com.github.se.cyrcle.model.user.UserViewModel
 import com.github.se.cyrcle.ui.navigation.NavigationActions
 import com.github.se.cyrcle.ui.navigation.Route
@@ -48,12 +50,8 @@ class ViewProfileScreenTest {
 
     val user =
         User(
-            userId = "1",
-            username = "janesmith",
-            firstName = "Jane",
-            lastName = "Smith",
-            email = "jane.smith@example.com",
-            profilePictureUrl = "http://example.com/jane.jpg")
+            UserPublic("1", "janesmith", "http://example.com/jane.jpg"),
+            UserDetails("Jane", "Smith", ""))
 
     userViewModel = UserViewModel(mockUserRepository, mockParkingRepository)
     parkingViewModel = ParkingViewModel(mockImageRepository, mockParkingRepository)
--- a/app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ViewProfileScreenTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ViewProfileScreenTest.kt
@@ -51,7 +51,7 @@ class ViewProfileScreenTest {
     val user =
         User(
             UserPublic("1", "janesmith", "http://example.com/jane.jpg"),
-            UserDetails("Jane", "Smith", ""))
+            UserDetails("Jane", "Smith", "jane.smith@example.com"))
 
     userViewModel = UserViewModel(mockUserRepository, mockParkingRepository)
     parkingViewModel = ParkingViewModel(mockImageRepository, mockParkingRepository)
--- a/app/src/main/java/com/github/se/cyrcle/model/user/User.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/user/User.kt
@@ -11,16 +11,25 @@ package com.github.se.cyrcle.model.user
  * @property profilePictureUrl URL of the profile picture of the user.
  * @property favoriteParkings List of unique identifiers of the user's favorite parkings.
  */
-data class User(
+data class UserPublic(
     val userId: String,
     val username: String,
+    val profilePictureUrl: String = "",
+    // val lastLoginTime: Timestamp? = null,
+    // val accountCreationDate: Timestamp? = null,
+    // val numberOfContributedSpots: Int = 0,
+    // val userReputationScore: Int = 0
+)
+
+data class UserDetails(
     val firstName: String = "",
     val lastName: String = "",
-    val email: String,
-    val profilePictureUrl: String = "",
+    val email: String = "",
     val favoriteParkings: List<String> = emptyList(),
     // val lastLoginTime: Timestamp? = null,
     // val accountCreationDate: Timestamp? = null,
     // val numberOfContributedSpots: Int = 0,
     // val userReputationScore: Int = 0
 )
+
+data class User(val public: UserPublic, val details: UserDetails?)
--- a/app/src/main/java/com/github/se/cyrcle/model/user/User.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/user/User.kt
@@ -1,35 +1,42 @@
 package com.github.se.cyrcle.model.user
 
 /**
- * Data class representing a user.
+ * Data class representing the public information of a user.
  *
- * @property userId Unique identifier of the user.
- * @property username Username of the user.
- * @property firstName First name of the user.
- * @property lastName Last name of the user.
- * @property email Email of the user.
- * @property profilePictureUrl URL of the profile picture of the user.
- * @property favoriteParkings List of unique identifiers of the user's favorite parkings.
+ * @property userId The unique identifier of the user.
+ * @property username The username of the user.
+ * @property profilePictureUrl The path to the profile picture of the user on the cloud storage.
+ *   (Not implemented)
  */
 data class UserPublic(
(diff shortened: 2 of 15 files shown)


sean-prz:
@matttsss Could you please double check the firestore rules, even If you're not reviewing the whole code.

app/src/main/java/com/github/se/cyrcle/model/user/User.kt
```diff
@@ -11,16 +11,25 @@ package com.github.se.cyrcle.model.user
  * @property profilePictureUrl URL of the profile picture of the user.
  * @property favoriteParkings List of unique identifiers of the user's favorite parkings.
  */
-data class User(
+data class UserPublic(
```
Pijukai, line 14: Don't forget to update documentation here

app/src/main/java/com/github/se/cyrcle/model/user/User.kt
```diff
@@ -11,16 +11,25 @@ package com.github.se.cyrcle.model.user
...
     val userId: String,
     val username: String,
+    val profilePictureUrl: String = "",
+    // val lastLoginTime: Timestamp? = null,
```
Pijukai, line 18: I think that `lastLoginTime` should be in `UserDetails`. `numberOfContributionsSpots` and  `userReputationScore` should be public since we want to create a community driven app. Those are both stats that should be displayed to the community since they give a hint about how much credibility a review has based on the user that published it.

app/src/main/java/com/github/se/cyrcle/model/user/UserRepositoryFirestore.kt
```diff
@@ -52,30 +57,62 @@ constructor(private val db: FirebaseFirestore, private val auth: FirebaseAuth) :
...
+            }
+            .addOnFailureListener { onFailure(it) }
+      }
+    }
   }
 
   override fun updateUser(user: User, onSuccess: () -> Unit, onFailure: (Exception) -> Unit) {
+    Log.d("UserRepositoryFirestore", "updateUser: $user")
+    if (user.details == null) {
+      Log.d("UserRepositoryFirestore", "updateUser: user details are null, can't update")
```
Pijukai, line 100: Since we throw an exception here , is it necessary to keep this log.d ?

app/src/main/java/com/github/se/cyrcle/ui/authentication/AuthenticatorImpl.kt
```diff
@@ -91,9 +93,7 @@ class AuthenticatorImpl @Inject constructor(private val auth: FirebaseAuth) : Au
 
           val user =
               User(
-                  userId = authResult.user?.uid ?: "",
-                  username = authResult.user?.displayName ?: "",
-                  email = authResult.user?.email ?: "")
+                  UserPublic(authResult.user!!.uid, authResult.user!!.displayName!!), UserDetails())
```
Pijukai, line 96: Why did you removed the email argument here ? That's just to understand

app/src/test/java/com/github/se/cyrcle/model/user/UserTest.kt
```diff
@@ -5,17 +5,6 @@ import org.junit.Assert.assertNotEquals
 import org.junit.Test
 
 class UserTest {
-  @Test
-  fun testConstructor() {
```
Pijukai, line 9: Why did you removed this test ? Seems relevant to me.

app/src/main/java/com/github/se/cyrcle/ui/authentication/AuthenticatorImpl.kt
sean-prz: That is a big mistake on my part, thanks for spotting it

app/src/test/java/com/github/se/cyrcle/model/user/UserTest.kt
sean-prz: Testing the default constructor of a class sounds really useless. That is up to the dev Team of Kotlin.

app/src/main/java/com/github/se/cyrcle/model/user/User.kt
sean-prz: Will do, and add more documentation for your other comment about doc.

app/src/main/java/com/github/se/cyrcle/model/user/UserRepositoryFirestore.kt
sean-prz: No I forgot to remove my debugging log

app/src/main/java/com/github/se/cyrcle/model/user/UserRepositoryFirestore.kt
```diff
@@ -52,30 +57,62 @@ constructor(private val db: FirebaseFirestore, private val auth: FirebaseAuth) :
...
-                document.data?.let { deserializeUser(it) }
+                val userPublic = deserializeUser(document.data!!)
+                User(userPublic!!, null)
               }
           onSuccess(users)
         }
         .addOnFailureListener { onFailure(it) }
   }
 
   override fun addUser(user: User, onSuccess: () -> Unit, onFailure: (Exception) -> Unit) {
```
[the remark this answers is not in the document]
sean-prz: The documentation in inherited from the interface, and I don't change the signature of the function. 
It's ok to leave as it is ?
image

app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ViewProfileScreenTest.kt
```diff
@@ -45,12 +47,8 @@ class ViewProfileScreenTest {
...
     val user =
         User(
-            userId = "1",
-            username = "janesmith",
-            firstName = "Jane",
-            lastName = "Smith",
-            email = "jane.smith@example.com",
-            profilePictureUrl = "http://example.com/jane.jpg")
+            UserPublic("1", "janesmith", "http://example.com/jane.jpg"),
+            UserDetails("Jane", "Smith", ""))
```
Pijukai, line 51: Jane's Email isn't added to the `UserDetails`

app/src/main/java/com/github/se/cyrcle/model/user/UserRepositoryFirestore.kt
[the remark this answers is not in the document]
Pijukai: Ohhh I didn't see it my bad. Yes it's good to leave it as it is !

app/src/main/java/com/github/se/cyrcle/model/user/UserRepositoryFirestore.kt
```diff
@@ -32,12 +30,18 @@ constructor(private val db: FirebaseFirestore, private val auth: FirebaseAuth) :
...
+          db.collection(collectionPath)
+              .document(userId)
+              .collection("private")
+              .document("details")
+              .get()
+              .addOnSuccessListener { detailsDocument ->
+                val userDetails = deserializeUserDetails(detailsDocument.data!!)
+                onSuccess(User(userPublic, userDetails))
+              }
+          onSuccess(User(userPublic, null))
```
matttsss, line 44: Isn't this line going to always be executed? Since the `onSuccess` call two lines above will be called asynchronously. One solution would be to add this line in the `onFailure` listener of the second database call don't you think?

app/src/main/java/com/github/se/cyrcle/model/user/UserRepositoryFirestore.kt
```diff
@@ -52,30 +56,60 @@ constructor(private val db: FirebaseFirestore, private val auth: FirebaseAuth) :
...
-    db.collection(collectionPath)
-        .document(user.userId)
-        .set(serializeUser(user))
-        .addOnSuccessListener { onSuccess() }
-        .addOnFailureListener { onFailure(it) }
+    if (user.details == null) {
+      onFailure(Exception("User details are required"))
+      return
+    }
+    db.collection(collectionPath).document(user.public.userId).get().addOnSuccessListener {
```
matttsss, line 74: Please don't forget to add the onFailure listener to this database call as well!

matttsss, line 74: Please don't forget to add the onFailure listener here as well

app/src/main/java/com/github/se/cyrcle/model/user/UserRepositoryFirestore.kt
sean-prz: Your implementation sounds better than mine. This way we don't call onSuccess twice

Pijukai (approved):
Everything is good for me well done !!!! This will allow us to enhance privacy for the user. You just need to check that the CI passes now.

matttsss (approved):
All good for me! Thanks for the changes!

PR #179, opened 2024-11-11
Text input verification

Armetiis:
### Key changes
Introduced a new atom component called ConditionCheckingInputText. This component:
- Wraps the existing InputText component
- Validates input's length against given bounds
- Displays an appropriate error message for invalid input

The InputText component has been enhanced with the following features:
- isError argument: Highlights the input field in red if validation fails
- hasClearIcon argument: When true, displays a button to clear the user's input

The new ConditionCheckingInputText has been integrated into relevant screens
"Save" and "Submit" buttons are now disabled when input validation fails
The affected screens are
- AttributePicker
- EditProfile
- AddReview

No error vs with error:

image
image

### Coverage
image

### DIFF
--- a/app/src/androidTest/java/com/github/se/cyrcle/ui/addParking/AddScreensNavigationTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/ui/addParking/AddScreensNavigationTest.kt
@@ -11,6 +11,7 @@ import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithTag
 import androidx.compose.ui.test.onNodeWithText
 import androidx.compose.ui.test.performClick
+import androidx.compose.ui.test.performTextInput
 import androidx.lifecycle.viewmodel.compose.viewModel
 import androidx.navigation.compose.rememberNavController
 import androidx.test.ext.junit.runners.AndroidJUnit4
@@ -66,7 +67,7 @@ class AddScreensNavigationTest {
         addressViewModel,
         mockedAuthenticator,
         LocalContext.current as android.app.Activity)
-    return listOf<Any>(
+    return listOf(
         navigationActions,
         parkingViewModel,
         reviewViewModel,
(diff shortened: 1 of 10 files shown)


app/src/androidTest/java/com/github/se/cyrcle/ui/addParking/AddScreensNavigationTest.kt
```diff
@@ -66,7 +67,7 @@ class AddScreensNavigationTest {
         addressViewModel,
         mockedAuthenticator,
         LocalContext.current as android.app.Activity)
-    return listOf<Any>(
+    return listOf(
```
LouisJamesVasseur, line 70: Does this change anything in practice?

app/src/androidTest/java/com/github/se/cyrcle/ui/review/ReviewScreenTest.kt
```diff
@@ -75,7 +75,7 @@ class ReviewScreenTest {
     composeTestRule
         .onNodeWithTag("TopAppBarTitle")
         .assertIsDisplayed()
-        .assertTextContains("Add Your Review")
+        .assertTextContains("Add your review")
```
LouisJamesVasseur, line 78: Why remove the captialization? It's a title...

LouisJamesVasseur (approved):
From both a UI and code perspective, this is a clear and robust implementation. Well done!

app/src/androidTest/java/com/github/se/cyrcle/ui/review/ReviewScreenTest.kt
Armetiis: It's true. I just tried removing some inconsistent caps in the strings ressources without really overthinking it. Do you really think it's ok to have "Add Your Review"? To me it looks illegal. Anyways we can come back to these details later, let's keep moving forward witht the important stuff!

LouisJamesVasseur: Completely agree!

PR #182, opened 2024-11-11
Implement Focus Mode

Pijukai:
## Code Coverage

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
@@ -3,11 +3,13 @@ package com.github.se.cyrcle.ui.map
 import android.app.Activity
 import android.graphics.Bitmap
 import android.graphics.BitmapFactory
+import android.util.Log
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.Add
+import androidx.compose.material.icons.filled.MyLocation
 import androidx.compose.material3.Scaffold
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
@@ -47,6 +49,7 @@ import com.mapbox.geojson.Polygon
 import com.mapbox.maps.CameraBoundsOptions
 import com.mapbox.maps.CameraOptions
 import com.mapbox.maps.CameraState
+import com.mapbox.maps.EdgeInsets
 import com.mapbox.maps.MapIdleCallback
 import com.mapbox.maps.MapView
 import com.mapbox.maps.ScreenCoordinate
@@ -68,6 +71,10 @@ import com.mapbox.maps.plugin.gestures.gestures
 import com.mapbox.maps.plugin.locationcomponent.OnIndicatorPositionChangedListener
 import com.mapbox.maps.plugin.locationcomponent.createDefault2DPuck
 import com.mapbox.maps.plugin.locationcomponent.location
+import com.mapbox.maps.plugin.viewport.data.FollowPuckViewportStateOptions
+import com.mapbox.maps.plugin.viewport.data.OverviewViewportStateOptions
+import com.mapbox.maps.plugin.viewport.state.FollowPuckViewportState
+import com.mapbox.maps.plugin.viewport.state.ViewportState
 import com.mapbox.maps.viewannotation.annotatedLayerFeature
 import com.mapbox.maps.viewannotation.annotationAnchor
 import com.mapbox.maps.viewannotation.geometry
@@ -101,7 +108,12 @@ fun MapScreen(
 
   val screenCapacityString = stringResource(R.string.map_screen_capacity)
 
-  val bitmap = BitmapFactory.decodeResource(LocalContext.current.resources, R.drawable.red_marker)
+    val focusMode = remember { mutableStateOf(false) }
+
+
+
+
+    val bitmap = BitmapFactory.decodeResource(LocalContext.current.resources, R.drawable.red_marker)
   val resizedBitmap = Bitmap.createScaledBitmap(bitmap, 100, 150, false)
   // Draw markers on the map when the list of parkings changes
   LaunchedEffect(listOfParkings, pointAnnotationManager, selectedParking?.nbReviews) {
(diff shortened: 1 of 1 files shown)


sean-prz (commented):
I feel like the design choice is a bit awkward. The standard for google maps and others it to have the map go back to the free mode when you scroll away, not to lock you in the focus mode. Can't we just add a listener for scroll of the user to call updateFocusMode(false) ? 

Also you're missing your code coverage. I'll check out the code and try on my phone in the next hour

app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
```diff
@@ -78,6 +96,61 @@ class MapViewModel : ViewModel() {
     _selectedLocation.value = location
   }
 
+  /**
+   * h Get the bottom left and top right corners of the screen in latitude and longitude
```
andrepeiry, line 100: Possible typo with the front-facing `h`

app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
```diff
@@ -78,6 +96,61 @@ class MapViewModel : ViewModel() {
...
+      this.enabled = true
+      this.locationPuck = createDefault2DPuck(true)
+    }
+    val onIndicatorPositionChangedListener = OnIndicatorPositionChangedListener { location ->
+      mapView.mapboxMap.setCamera(CameraOptions.Builder().center(location).build())
+      mapViewModel.updateCameraPosition(mapView.mapboxMap.cameraState)
+    }
+
+    locationComponentPlugin.addOnIndicatorPositionChangedListener(
+        onIndicatorPositionChangedListener)
```
andrepeiry, line 151: [Non-blocking note] In general, I would avoid having a such intermediary variable, as they tend to make the code harder to read. In my opinion, it would have been fine to go directly for
```
locationComponentPlugin.addOnIndicatorPositionChangedListener( { location ->
        mapView.mapboxMap.setCamera(CameraOptions.Builder().center(location.build())
        mapViewModel.updateCameraPosition(mapView.mapboxMap.camera)
    }
)

**Comment by Pijukai** [comment]
> I feel like the design choice is a bit awkward. The standard for google maps and others it to have the map go back to the free mode when you scroll away, not to lock you in the focus mode. Can't we just add a listener for scroll of the user to call updateFocusMode(false) ?
> 
> Also you're missing your code coverage. I'll check out the code and try on my phone in the next hour

I can try to correct that but unfortunately it is not supported by MapBox library internally which makes thing a bit difficult.

File: app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt (hunk 2 above)
Comment (Pijukai) [reply app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt]: fixed thanks

File: app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt (hunk 1 above)
Comment (Pijukai) [reply app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt]: fixed

File: app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt [hunk 3]
```diff
@@ -24,6 +29,8 @@ class MapViewModel : ViewModel() {
   private val _locationPickerState = MutableStateFlow(LocationPickerState.NONE_SET)
   val locationPickerState: StateFlow<LocationPickerState> = _locationPickerState
 
+  private val _focusMode = MutableStateFlow(false)
+  val focusMode: StateFlow<Boolean> = _focusMode
```
andrepeiry, line 33: I don't think `focusMode` is a good name for a boolean. What does `true` and `false` mean ? 
- The easy solution is to rename it as `isTrackingModeEnable` (or something similar). 
- The more engineered solution (often overkill) is to create an `Enum` class that contains all the possibilities, and you can keep the name

sean-prz (changes requested):
I launched the app to try the feature and, the behaviour described in the PR doesn't work for me at all. I can still scroll off in the two modes.

Note that this is exactly the behaviour I wanted and I mentioned in my previous comment. 

Finally before you re-request me for a second review, make the two buttons symmetrical, this is painful
image

Pijukai:
It's my fault for not being really clear. For now, we can still move the camera as we want in the two modes but that can creates unexpected behavior. For example if we scroll while being in FocusMode state then it is going against what the state was created for, the free state being OverviewState.  That's why it is for me more intuitive to lock the user while in FocusMode.

app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
```diff
@@ -45,6 +52,16 @@ class MapViewModel : ViewModel() {
...
+  /**
+   * Update the focus mode
+   *
+   * @param focusMode the new focus mode This function is used to update the focus mode of the map
+   *   screen. The focus mode is used to center the map on the user's location
+   */
+  fun updateFocusMode(focusMode: Boolean) {
+    _focusMode.value = focusMode
+  }
+
```
andrepeiry, line 64: A `toggleFocusMode()` function would be welcome, as most of the time, we don't want to bother with actually updating, rather just switching the value.

andrepeiry (changes requested):
Great job making the map following the user's location !
Make sure not to make your code more complicated that it need to be ^^ It is often said that the two most difficult things in Tech are cache coherency and naming things, and understandably. Please refer to in-file comment on how to improve them (some are non-blocking, don't waste time on them if you don't have too much)

app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
Pijukai: I chose this name to stay consisten with the rest of the function of the mapViewModel

Pijukai:
I am sorry for the wait but i have finally corrected all the problems you brought up and finally reached the behavior you wanted Sean which is the correct one.

app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
Pijukai: I changed to your first proposition , the second one being as you said a bit overkill. Thanks for the propositions !

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -252,44 +285,33 @@ fun MapScreen(
...
+          onClick = {
+            mapViewModel.updateTrackingMode(!mapViewModel.isTrackingModeEnable.value)
+
+            mapViewportState.transitionToFollowPuckState(
+                FollowPuckViewportStateOptions.Builder()
+                    .pitch(0.0)
+                    .zoom(maxZoom)
+                    .padding(EdgeInsets(500.0, 100.0, 100.0, 100.0))
+                    .build())
+            mapViewModel.updateTrackingMode(true)
```
sean-prz, line 306: updateTrackingMode is being called twice in the OnClick is it intended ?

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -252,44 +285,33 @@ fun MapScreen(
...
+                  .scale(1.2f)
+                  .testTag("recenterButton"),
+          onClick = {
+            mapViewModel.updateTrackingMode(!mapViewModel.isTrackingModeEnable.value)
+
+            mapViewportState.transitionToFollowPuckState(
+                FollowPuckViewportStateOptions.Builder()
+                    .pitch(0.0)
+                    .zoom(maxZoom)
+                    .padding(EdgeInsets(500.0, 100.0, 100.0, 100.0))
```
sean-prz, line 304: There is weird "glitch"/"shake" when we go off the focus mode. 
I think it's due to this Insets being 500 while all other one are 100.

I tried modifying it and it seems to resolve the shake

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
sean-prz: I tried removing the first call, and the functionnality is preserved

sean-prz (approved):
Appart from the edgeInsets causing some rollback, 
I am very happy with this feature. Huge improvement between the two versions you submitted for review, Thanks for taking the time to make things smooth and pretty

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Pijukai: Okay i will change that thanks !

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Pijukai: No it is not i will remove it right away !

andrepeiry (approved):
Thank you for you work and adaptation. Everything seems to be in order now.

Issue #184, opened 2024-11-11
Adding a review does not update the user's average rating of the spot.


sean-prz:
### Problem
When a user adds a review and come back to the detail screen, the average score and number of review are not updated.
This is because informations are cached, and becomes stale.

### Solution
make sure adding a review also updates the cache and not only the database

Issue #183, opened 2024-11-11
Reviews required to navigate twice to the screen to be displayed


sean-prz:
### Problem
When clicking on _see all review_ the review are not shown on the screen.
The user needs to navigate back to the previous screen and then click once again on _see all review_ to see them

Issue #180, opened 2024-11-11
All reviews are displayed as if the current user made all of them 


sean-prz:
#### Problem
When on the AllReviewsScreen, the author of the review is not the correct one, instead it is the currently logged-in user who's name is displayed.

PR #185, opened 2024-11-11
Ensure synchronisation between the Viewmodel and the UI

sean-prz:
_closes #180_
_closes #183_
_closes #184_

### This PR addresses multiples bugs causing the UI to be out of sync with the data.
See the issue for more information about the issue that it fixes. 

### Fixing #180
A state has been introduced that holds the string of the username of the review's owner.
This state has a default value of "Unknown user" and is updated when the backend is done fetching the username.
The review depend on this state, hence it is recomposed when the request terminates. 

This fix also prevent the viewmodel from overwritting the currentUser state, which holds the logged-in user, with the result of the query. This is done by overloading the function of the viewmodel with another version that does not replace the currentUser value

### Fixing #183 
Lifting the state outside the lazyColomn fixes this issue. 
The lazyColomn is now recomposed every time the state is updated.

### Fixing #184
To fix this, the function of the viewmodel that updatesTheReviewScore now also update the cache and not only the database.
Keeping the cache in sync with the data.

-- - 
image

### DIFF
--- a/app/src/androidTest/java/com/github/se/cyrcle/ui/review/AllReviewsScreenTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/ui/review/AllReviewsScreenTest.kt
@@ -32,11 +32,11 @@ class AllReviewsScreenTest {
 
   @Before
   fun setUp() {
-    parkingViewModel.selectParking(TestInstancesParking.parking2)
     reviewViewModel.addReview(TestInstancesReview.review4)
     reviewViewModel.addReview(TestInstancesReview.review3)
     reviewViewModel.addReview(TestInstancesReview.review2)
     reviewViewModel.addReview(TestInstancesReview.review1)
+    parkingViewModel.selectParking(TestInstancesParking.parking2)
   }
 
   @Test
@@ -45,7 +45,6 @@ class AllReviewsScreenTest {
     composeTestRule.setContent {
       AllReviewsScreen(navigationActions, parkingViewModel, reviewViewModel, userViewModel)
       userViewModel.setCurrentUser(TestInstancesUser.user1)
-      parkingViewModel.selectParking(TestInstancesParking.parking1)
     }
 
     composeTestRule.onNodeWithTag("AllReviewsScreenBox").assertIsDisplayed()
@@ -57,8 +56,6 @@ class AllReviewsScreenTest {
   fun clickingReviewCard_expandsAndCollapsesCard() {
     composeTestRule.setContent {
       AllReviewsScreen(navigationActions, parkingViewModel, reviewViewModel, userViewModel)
-      userViewModel.setCurrentUser(TestInstancesUser.user1)
-      parkingViewModel.selectParking(TestInstancesParking.parking1)
     }
 
     composeTestRule.onNodeWithTag("ReviewCard0").performClick()
@@ -75,7 +72,6 @@ class AllReviewsScreenTest {
     composeTestRule.setContent {
       AllReviewsScreen(navigationActions, parkingViewModel, reviewViewModel, userViewModel)
       userViewModel.setCurrentUser(TestInstancesUser.user1)
-      parkingViewModel.selectParking(TestInstancesParking.parking1)
     }
 
     composeTestRule.onNodeWithTag("ReviewCard1").performClick()
--- a/app/src/main/java/com/github/se/cyrcle/model/user/UserRepositoryFirestore.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/user/UserRepositoryFirestore.kt
@@ -1,5 +1,6 @@
 package com.github.se.cyrcle.model.user
 
+import android.util.Log
 import com.google.firebase.Firebase
 import com.google.firebase.auth.FirebaseAuth
 import com.google.firebase.auth.auth
@@ -30,7 +31,15 @@ constructor(private val db: FirebaseFirestore, private val auth: FirebaseAuth) :
         .document(userId)
         .get()
         .addOnSuccessListener { document ->
-          val userPublic = deserializeUser(document.data!!)
+          val userPublic: UserPublic =
+              try {
+                Log.d("UserRepositoryFirestore", " Correctly fetched UserPublic")
+                deserializeUserPublic(document.data!!)
+              } catch (e: Exception) {
+                Log.d("UserRepositoryFirestore", "Error fetching UserPublic")
+                onFailure(e)
+                return@addOnSuccessListener
+              }
 
           db.collection(collectionPath)
               .document(userId)
@@ -41,7 +50,7 @@ constructor(private val db: FirebaseFirestore, private val auth: FirebaseAuth) :
                 val userDetails = deserializeUserDetails(detailsDocument.data!!)
                 onSuccess(User(userPublic, userDetails))
               }
-              .addOnFailureListener(onFailure)
+              .addOnFailureListener { onSuccess(User(userPublic, null)) }
         }
         .addOnFailureListener(onFailure)
   }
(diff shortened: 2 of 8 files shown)


app/src/androidTest/java/com/github/se/cyrcle/ui/review/AllReviewsScreenTest.kt
```diff
@@ -32,11 +32,11 @@ class AllReviewsScreenTest {
 
   @Before
   fun setUp() {
-    parkingViewModel.selectParking(TestInstancesParking.parking2)
     reviewViewModel.addReview(TestInstancesReview.review4)
     reviewViewModel.addReview(TestInstancesReview.review3)
     reviewViewModel.addReview(TestInstancesReview.review2)
     reviewViewModel.addReview(TestInstancesReview.review1)
+    parkingViewModel.selectParking(TestInstancesParking.parking2)
```
LouisJamesVasseur, line 39: Is this part of the changes that fixes the bugs? If so, do you have an idea which one and why?

app/src/main/java/com/github/se/cyrcle/ui/review/AllReviewsScreen.kt
```diff
@@ -57,27 +61,30 @@ fun AllReviewsScreen(
...
-      if (userViewModel.currentUser.value?.public?.userId != null) {
-        reviewViewModel.parkingReviews.value.any {
-          it.owner == userViewModel.currentUser.value?.public?.userId
-        }
-      } else {
-        false
-      }
+  // When the screen is composed, clear the reviews and fetch the reviews for the selected parking
+  LaunchedEffect(Unit) {
+    reviewViewModel.clearReviews()
```
LouisJamesVasseur, line 72: why do the reviews need to be cleared first?

app/src/main/java/com/github/se/cyrcle/ui/review/AllReviewsScreen.kt
```diff
@@ -94,24 +101,25 @@ fun AllReviewsScreen(
...
-                            userViewModel.getUserById(curReview.owner)
-                            val uidOfOwner =
-                                userViewModel.currentUser.value?.public?.username ?: "none"
-                            if (index == selectedCardIndex) {
+                          items(items = reviewsList) { curReview ->
+                            val index = reviewsList.indexOf(curReview)
+                            val defaultUsername = stringResource(R.string.undefined_username)
+                            val uidOfOwner = remember { mutableStateOf(defaultUsername) }
+                            userViewModel.getUserById(curReview.owner) {
+                              uidOfOwner.value = it.public.username
```
LouisJamesVasseur, line 109: just to verify w.r.t edge cases, is it possible that the object here may be null? (since error handling for null case was handled in the original code)

LouisJamesVasseur (approved):
Great work on all the fixes! The code fixes are very smart (solidarizing w/ the map,  the different definitions of some objects which now make a lot more sense...), it's clear a lot of work went into them and all the bugs are well dealt with.

app/src/main/java/com/github/se/cyrcle/ui/review/AllReviewsScreen.kt
sean-prz: In the scenario where a user first look at the review of spot1 and then look at the reviews of spot2.
The user would still see the reviews of spot1 while the reviews of spot2 are fetched from the db

app/src/main/java/com/github/se/cyrcle/ui/review/AllReviewsScreen.kt
sean-prz: yes, if the viewmodel getUserById fails because of a null arg, or because the owner does not exist anymore, it will call the onFailure, which is not defined, so nothing happens

LouisJamesVasseur: Figured as much, good to see this was thought of and handled correctly

app/src/main/java/com/github/se/cyrcle/ui/review/AllReviewsScreen.kt
LouisJamesVasseur: Thanks for the example!

app/src/androidTest/java/com/github/se/cyrcle/ui/review/AllReviewsScreenTest.kt
sean-prz: If you mean moving this call to after the .addReview, then it makes no difference. I was trying to understand why the tests didn't pass, but turns out it wasn't because of this, it was because in the test we set the selected Parking to testInstances1 while all the test instances review are linked to the testinstances2.

TLDR : I didn't have to change this.

LouisJamesVasseur: That's what I had guessed, wanted to ensure that our implementation's functioning didn't depend on the order we did things which would be quite fragile in my opinion

app/src/main/java/com/github/se/cyrcle/model/user/UserRepositoryFirestore.kt
```diff
@@ -41,7 +50,7 @@ constructor(private val db: FirebaseFirestore, private val auth: FirebaseAuth) :
                 val userDetails = deserializeUserDetails(detailsDocument.data!!)
                 onSuccess(User(userPublic, userDetails))
               }
-              .addOnFailureListener(onFailure)
+              .addOnFailureListener { onSuccess(User(userPublic, null)) }
```
andrepeiry, line 53: It feels weird to have an `onSuccess` for an `FailureListener`. If it is an error, please fix, but if it is intended, it would be nice to have a small comment asserting that this is expected.

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
```diff
@@ -54,6 +54,25 @@ class UserViewModel(
...
+        userId,
+        onSuccess = { onSuccess(it) },
+        onFailure = { exception ->
+          Log.e(
+              "com.github.se.cyrcle.model.user.UserViewModel",
+              "Failed to fetch user by ID: $userId",
+              exception)
+        })
+  }
+
```
andrepeiry, line 75: I have seen no test for this new function. Would it be possible to have at least a small one ?

[Non-blocking note] Also, the tags for logging (`"UserViewModel"` and `"com.github.se.cyrcle.model.user.UserViewModel"`) don't match, maybe it's intended (as both are for different purposes)

andrepeiry (changes requested):
Nice work for fixing these bugs !

I have left some in-file comments, I'll let you take a look at them ^^

app/src/main/java/com/github/se/cyrcle/model/user/UserRepositoryFirestore.kt
sean-prz: It's very much intended. We let the database deny our request if we don't have access instead of checking in viewmodel if we have permission. 

This is more of a implementation choice. I will add a comment to explain this reasoning

sean-prz:
Good remarks, what about now ?

app/src/main/java/com/github/se/cyrcle/model/user/UserRepositoryFirestore.kt
andrepeiry: Thank you for the clarification

andrepeiry (approved):
Thank you for the test and clarification. Everything looks fine to me !

PR #191, opened 2024-11-13
fix : back stack correctly clears itself.

sean-prz:
_closes #166_
### This small PR is about navigating in the app.
Its goal is the following  : 
- Not let any buttons without interactions.
- Clears the back stack when navigating to a TLD
- Propose the guest user to navigate back to the sign-in screen when on his profile screen.
- Make the navigation graph more readable and maintainable, keep the screen name coherent with the composable name

### Important changes
- Map, list and profile are now correctly set up as TLD so making a back action on them makes you leave the app.

### Other minors changes

- The popup() function has been modified to completely clear the back stack.
- The add button in guest mode is not shown. 
- The guest user now has the possibility to sign in
- The TopBar of the viewprofile screen has been removed because it contained a back arrow, but it is a TLD.
- REVIEW screen has been renamed ADD_REVIEW 
- CARD screen has been renamed PARKING_DETAILS
- Screens that belong to a common theme are placed in a ROUTE for clarity and readability.
- MAP, LIST and PARKING_DETAILS are placed alone in their ROUTE. 

### Why
Fixing the navigation brings our navigation closer to a standard android app.
the Renaming and reordering is for maintainability. 
The add Button removed and the profile screen of guest user is not let any button/screen with no interactions.

### Code coverage

image

### DIFF
--- a/app/src/androidTest/java/com/github/se/cyrcle/ui/testEnd2End/MainActivityTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/ui/testEnd2End/MainActivityTest.kt
@@ -241,7 +241,6 @@ class MainActivityTest {
 
     fun assertMapScreen() {
       composeTestRule.onNodeWithTag("MapScreen").assertIsDisplayed()
-      composeTestRule.onNodeWithTag("addButton").assertIsDisplayed().assertHasClickAction()
       composeTestRule.onNodeWithTag("ZoomControlsIn").assertIsDisplayed().assertHasClickAction()
       composeTestRule.onNodeWithTag("ZoomControlsOut").assertIsDisplayed().assertHasClickAction()
       composeTestRule.onNodeWithTag("NavigationBar").assertIsDisplayed()
--- a/app/src/main/java/com/github/se/cyrcle/ui/profile/CreateProfileScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/profile/CreateProfileScreen.kt
@@ -19,43 +19,39 @@ import com.github.se.cyrcle.R
 import com.github.se.cyrcle.model.user.UserViewModel
 import com.github.se.cyrcle.ui.navigation.NavigationActions
 import com.github.se.cyrcle.ui.navigation.Route
-import com.github.se.cyrcle.ui.navigation.TopLevelDestination
-import com.github.se.cyrcle.ui.navigation.TopLevelDestinations
 import com.github.se.cyrcle.ui.theme.atoms.Button
 import com.github.se.cyrcle.ui.theme.atoms.Text
 import com.github.se.cyrcle.ui.theme.molecules.BottomNavigationBar
-import com.github.se.cyrcle.ui.theme.molecules.TopAppBar
 
 @Composable
 fun CreateProfileScreen(navigationActions: NavigationActions, userViewModel: UserViewModel) {
-    val screenTitle = stringResource(R.string.profile_screen_title)
+  val screenTitle = stringResource(R.string.profile_screen_title)
 
-    Scaffold(
-        modifier = Modifier.fillMaxSize().testTag("CreateProfileScreen"),
-        bottomBar =  { BottomNavigationBar(navigationActions, selectedItem = Route.PROFILE) },
-       ) { padding ->
-        Box(modifier = Modifier.padding(padding)) {
-            Column(
-                modifier = Modifier.fillMaxSize().padding(16.dp),
-                verticalArrangement = Arrangement.Center,
-                horizontalAlignment = Alignment.CenterHorizontally,
-            ) {
-                Text(
-                    "You need to be signed-in to modify your profile",
-                    style = MaterialTheme.typography.titleLarge
-                )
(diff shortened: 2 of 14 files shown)


app/src/androidTest/java/com/github/se/cyrcle/ui/testEnd2End/MainActivityTest.kt
```diff
@@ -241,7 +241,6 @@ class MainActivityTest {
 
     fun assertMapScreen() {
       composeTestRule.onNodeWithTag("MapScreen").assertIsDisplayed()
-      composeTestRule.onNodeWithTag("addButton").assertIsDisplayed().assertHasClickAction()
```
Pijukai, line 244: We should add as a task for next sprint to implement the sign in/guest aspects of the app in the tests so that instead of deleting this line we could create more relevant tests

app/src/main/java/com/github/se/cyrcle/ui/profile/CreateProfileScreen.kt
```diff
@@ -1,55 +1,57 @@
...
+            style = MaterialTheme.typography.titleLarge)
+        Spacer(modifier = Modifier.height(16.dp))
+        Text(
+            "Sign in now to add new spots, share reviews, and help fellow cyclists find the perfect parking space.")
+        Spacer(modifier = Modifier.height(8.dp))
+        Button(
+            text = "Go back to sign in",
+            onClick = {
+              // navigate to the route to not clear backstack
+              navigationActions.navigateTo(Route.AUTH)
```
matttsss, line 51: You probably wanted to use `navigationAction.navigateTo(TopLevelDestinations.AUTH)` since the go back button from AUTH after clicking here does not make us quit the app

sean-prz: Here I specifically don't use the TLD because it clears the back stack and so the user can't use the back action if he changes his mind and doesn't want to login anymore.

app/src/androidTest/java/com/github/se/cyrcle/ui/testEnd2End/MainActivityTest.kt
sean-prz: Here it's ok there is already another test that check the exact same assertion. Once with a user logged-out, once with a user logged-in

sean-prz: image

app/src/main/java/com/github/se/cyrcle/ui/profile/CreateProfileScreen.kt
Pijukai: Well a solution could be to do as matteo said and maybe add an `AlertDialog` composable that ask the user if he is certain about going back to the login screen.

matttsss: I could see why it's a valid approach, it just changes the behaviour of the screen in both scenarios but it's ok I guess.

matttsss (approved):
All good, some choices are up to debate but you did a good job of convincing me

app/src/main/java/com/github/se/cyrcle/ui/profile/CreateProfileScreen.kt
sean-prz: There is no need, this works perfectly right now : 
1. The guest user decides to sign in
2. He is redirected on the auth screen that is put on top of the stack instead of clearing it
3. This means now the user can back track, change his mind, use the go back and resume his session on the profile screen.
3.bis He logs in, is redirected to the map screen, the back stack is cleared. 

Am I missing something ? do you want the back stack to be cleared ? does not seem very conviniant to me

sean-prz:
> All good, some choices are up to debate but you did a good job of convincing me

I am of course open to discussion, we can always refactor the navigation later on.

app/src/androidTest/java/com/github/se/cyrcle/ui/testEnd2End/MainActivityTest.kt
Pijukai: I agree that functionally speaking the button is asserted correctly multiple times. However, we should try to make the E2E tests as complete and independant as possible (and here we do not assert that the button does not exist in the E2E tests). Again, this is just a minor detail that we can adress later. That's a non blocking issue for me.

app/src/main/java/com/github/se/cyrcle/ui/profile/CreateProfileScreen.kt
matttsss: No no it makes sense, the fact that the stack is cleared when navigating to the map solves my "grudge". And I still had a mixup between TLDs and screens but its good now

Pijukai:
By the way, because @matttsss and me didn't see it but could you add the coverage ? Then one of us will quickly approved again as you need a reapproval and you can move on

Pijukai (approved):
Thanks !

PR #192, opened 2024-11-13
Hide centering button when location is disabled

### LINKED ISSUE #187: The centering button on the map is displayed even when the position is disabled
The button for centering should only appear when our position is enabled and trackable

matttsss:
### What is this PR?
This PR aims to add a quickfix to hide the centering button when the location is disabled.
This could not simply be done using an if since it would have asked a lot of work changing the permission manager (for tests to pass), this could be done in a later sprint after M2.

### How?
It will be left as an exercise for the reviewer to find the button when location is disabled :)

### Test Coverage
image

Closes #187


Pijukai (approved):
It would be great to add a TODO to not forget to change this next sprint . Anyway, great easter egg despite the fact that this is not easter :+1:

sean-prz (approved):
A quick and ugly fix ! I agree let's make sure anyone reading the code know for sure this is temporary

PR #194, opened 2024-11-13
Toggle between Rectangles and Markers on map

sean-prz:
_closes #94_
### Toggle between simple and advance mode on map screen

- The user can now choose if he wants to see markers or more detailed rectangles. 
- The pin on the map has been changed to not be a default red marker anymore

### How

- introduce a state that holds the mode of the map.
- Mode of the map are represented by an enum added in this PR
- added a toggle switch that changes the map mode state.
- when the map is too unzoomed the state is also automatically changed and the switch hidden on the UI.
- the state is observed and  either call `drawMarkers` or `drawRectangles`

A good portion of the lines changed in this PR are just reordering and grouping lines of code in the map screen for more readability.
-- -
### Demo. 
https://github.com/user-attachments/assets/0cfdb076-c1e7-456a-86a8-4ade0861c7da
-- -
### Code coverage 
image

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
@@ -117,6 +117,8 @@ fun MapScreen(
   val enableParkingAddition by userViewModel.isSignedIn.collectAsState(false)
   // create a remember  state to store if the markers or the rectangles are displayed
   val mapMode = remember { mutableStateOf(MapMode.MARKERS) }
+  // this is the state the user selected by the user, remembered even when zoomed out.
+  val userMapMode = remember { mutableStateOf(MapMode.MARKERS) }
   val mapViewportState = MapConfig.createMapViewPortStateFromViewModel(mapViewModel)
   var removeViewAnnotation = remember { true }
   var cancelables = remember { Cancelable {} }
@@ -308,12 +310,19 @@ fun MapScreen(
                   if (mapView.mapboxMap.cameraState.zoom > thresholdDisplayZoom) {
                     parkingViewModel.getParkingsInRect(currentBottomLeft, currentTopRight)
                   }
+                  // On zoom-out past the threshold, switch to the markers mode
                   if (mapView.mapboxMap.cameraState.zoom < LABEL_THRESHOLD &&
                       zoomState.value >= LABEL_THRESHOLD) {
                     mapMode.value = MapMode.MARKERS
                   }
+                  // On zoomin in past the threshold, switch to the user's selected mode
+                  if (mapView.mapboxMap.cameraState.zoom >= LABEL_THRESHOLD &&
+                      zoomState.value < LABEL_THRESHOLD) {
+                    mapMode.value = userMapMode.value
+                  }
 
                   // store the zoom level
+                  // This must stay at the end of the listener.
                   zoomState.value = mapView.mapboxMap.cameraState.zoom
                 }
             // =======================  CAMERA  LISTENER  =======================
@@ -340,7 +349,8 @@ fun MapScreen(
             Switch(
                 checked = mapMode.value.isAdvancedMode,
                 onCheckedChange = {
-                  mapMode.value = if (it) MapMode.RECTANGLES else MapMode.MARKERS
+                  userMapMode.value = if (it) MapMode.RECTANGLES else MapMode.MARKERS
+                  mapMode.value = userMapMode.value
                 },
                 colors =
                     SwitchDefaults.colors()
--- a/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
@@ -4,6 +4,7 @@ import android.app.Activity
 import android.graphics.Bitmap
 import android.graphics.BitmapFactory
 import android.util.Log
+import androidx.compose.animation.core.animateFloatAsState
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxSize
@@ -64,7 +65,6 @@ import com.mapbox.maps.plugin.annotation.AnnotationSourceOptions
 import com.mapbox.maps.plugin.annotation.ClusterOptions
 import com.mapbox.maps.plugin.annotation.annotations
 import com.mapbox.maps.plugin.annotation.generated.OnPointAnnotationClickListener
-import com.mapbox.maps.plugin.annotation.generated.PointAnnotation
 import com.mapbox.maps.plugin.annotation.generated.PointAnnotationManager
 import com.mapbox.maps.plugin.annotation.generated.PointAnnotationOptions
 import com.mapbox.maps.plugin.annotation.generated.PolygonAnnotationManager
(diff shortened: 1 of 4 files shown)


AlexandreBeeDeeBop:
I havent read the code yet so maybe this is not your problem but here's a few things I noticed while on the emulator on your branch :

- The little button to toggle Advanced Mode could be improved by
1. having a different color if it's not toggled (maybe white) than when it's toggled (maybe blue like it is right now for both modes).
2. having the text above it be Normal mode when it's not toggled and Advanced mode when it's toggled (like it is right now for both modes).

- When you activate advanced mode -> dezoom so that "when the map is too unzoomed the state is also automatically changed and the switch hidden on the UI." as you say -> and zoom back, you are not in advanced mode anymore, which can be annoying.


That's it for now, the feature is really cool and now im going to see if I can find issues in the code itself.

sean-prz:
> I havent read the code yet so maybe this is not your problem but here's a few things I noticed while on the emulator on your branch :
> 
> * The little button to toggle Advanced Mode could be improved by
> 
> 1. having a different color if it's not toggled (maybe white) than when it's toggled (maybe blue like it is right now for both modes).
> 2. having the text above it be Normal mode when it's not toggled and Advanced mode when it's toggled (like it is right now for both modes).
> 
> * When you activate advanced mode -> dezoom so that "when the map is too unzoomed the state is also automatically changed and the switch hidden on the UI." as you say -> and zoom back :
>   you are not in advanced mode anymore, which can be annoying.
> 
> That's it for now, the feature is really cool and now im going to see if I can find issues in the code itself.

I totally with prop 1, but I am waiting for the theme to be defined, because I don't want to put any hardcoded value so for now it's just PrimaryColor which is the only one that is not ugly.

I don't really get point 3 can you rephrase ?

AlexandreBeeDeeBop:
Ok fair enough maybe we dont need to have the text be changed, but the color yes we agree on that.
Ill explain how you can test my points 3 :
Zoom to the max, activate advanced mode, dezoom to the max, zoom to the max. You are now in normal modem not in advanced mode anymore. I give you the extreme one so you see the problem, the thing is that even if you dont exaggerate by max zooming then max dezooming etc, there's still this one moment when a single click on the - followed by a single click on the + can make you go from advanced mode to normal mode, which might by annoying for users

sean-prz:
OK I understand now, I can try to add another state to remember the user selectionb

AlexandreBeeDeeBop:
Ok good, it's because of this feature you added isnt it ? :

"when the map is too unzoomed the state is also automatically changed and the switch hidden on the UI."

AlexandreBeeDeeBop:
ok, can you add a test that zooms, activates advanced mode, unzooms, zooms, asserts advanced mode is still activated ?

sean-prz:
Complicated, because every state is internal. I can extend the mapscreen test to assert that the toggle switch is there

AlexandreBeeDeeBop:
nevermind dont bother we can test it ourselves with the emulator. I was thinking about the test coverage.

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -249,10 +310,25 @@ fun MapScreen(
...
+
+                  // store the zoom level
+                  // This must stay at the end of the listener.
+                  zoomState.value = mapView.mapboxMap.cameraState.zoom
                 }
+            // =======================  CAMERA  LISTENER  =======================
 
             onDispose {
-              pointAnnotationManager?.deleteAll()
+              markerAnnotationManager?.deleteAll()
```
Armetiis, line 331: here shouldn't we delete all annotations of all kind? (rectangle and pLabel)

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -261,6 +337,38 @@ fun MapScreen(
 
     // ======================= OVERLAY =======================
     Box(modifier = Modifier.padding(padding).fillMaxSize()) {
+      // A switch to change the display mode
+      Column(
+          modifier =
+              Modifier.align(Alignment.TopStart).padding(start = 8.dp, top = 32.dp).alpha(alpha),
+          horizontalAlignment = Alignment.CenterHorizontally) {
+            Text(
+                "Advanced Mode",
```
Armetiis, line 346: Please use string resources

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -319,26 +418,40 @@ fun MapScreen(
...
       val polygonAnnotationOptions =
           PolygonAnnotationOptions()
               .withGeometry(polygon)
-              .withFillColor("#22799B")
+              .withFillColor("#1A4988")
               .withFillOpacity(0.7)
-      polygonAnnotationManager?.create(polygonAnnotationOptions)
+      annotations.add(polygonAnnotationOptions)
+      val area = TurfMeasurement.area(polygon)
+      Log.d("MapScreen", "Area: $area")
```
Armetiis, line 440: I think you forgot to remove the log here

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -71,12 +80,28 @@ import com.mapbox.maps.viewannotation.annotatedLayerFeature
...
 import com.mapbox.maps.viewannotation.geometry
 import com.mapbox.maps.viewannotation.viewAnnotationOptions
+import com.mapbox.turf.TurfMeasurement
 
 const val defaultZoom = 16.0
 const val maxZoom = 18.0
 const val minZoom = 8.0
 const val thresholdDisplayZoom = 13.0
 const val LAYER_ID = "0128"
+const val LABEL_THRESHOLD = 15.5
```
Armetiis, line 90: In my opinion this name is not very clear. Maybe something like "Advanced mode zoom treshold" would be clearer?

Armetiis (approved):
Really great job you did here. Everything looks great. 
Apart from the comments I left you in the code I have 2 suggestions

1. As Alexandre said, having the toggle switch gray when disabled would really improve it. Also, I think it would look better if the text and the toggle were one next to the other (in a row rather than column). You can remove "Mode" in "Advanced Mode" so the row isn't too long

2. I don't know how much work this might imply, so if you consider it too hard/long, this can be the subject of a future PR. Otherwise, you think you can make changes so that clicking on a parking rectangle does the same as clicking on a ping? (Display the preview card)

None of this is blocking, so you have my approval. However, these little changes could make a difference. Up to you.
Also try to update your coverage after rebasing

PS: Thank you very much for taking the time to refactor the file

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
sean-prz: Fixed

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
sean-prz: Fixed

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
sean-prz: Fixed

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
sean-prz: Fixed

sean-prz:
Thanks for all the very pertinent comments, I fixed them all.

Armetiis:
Prop 1 is perfect!

AlexandreBeeDeeBop (approved):
Once you're done adressing all the changes requested you're good to merge.

PR #195, opened 2024-11-13
List screen based on user's actual location

### LINKED ISSUE #196: Use the user location as a reference point in the ListScreen

Armetiis:
This PR updates the list screen to use the user's current location as the reference point for displaying the closest parkings, replacing the previous hardcoded point.

### Implementation Details:
- Used MapBox MapView to access the location component plugin
- Implemented a location listener with a 1Hz update frequency using a debouncer
- Use hardcoded point `EPFLCenter` if location permission is disabled

### Coverage
Coverage as of 13/11 19:15
image

-- -
Closes #196

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
@@ -159,12 +159,12 @@ class MapViewModel : ViewModel() {
 
   private val handler = Handler(Looper.getMainLooper())
   private var lastUpdateTime = 0L
-  private val updateInterval = 1000L // 1 second
+  private val updateInterval = 500L // 0.5 seconds
 
   /**
    * Initialize the location component of the map and add a listener to update the user position
    *
-   * @param locationComponentPlugin the location component plugin to initialize
+   * @param mapView the MapView to initialize the location component on
    */
   fun initLocationComponent(mapView: MapView) {
     val locationComponentPlugin = mapView.location
--- a/app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
@@ -1,9 +1,12 @@
 package com.github.se.cyrcle.model.map
 
+import android.os.Handler
+import android.os.Looper
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.ViewModelProvider
 import com.github.se.cyrcle.model.parking.Location
 import com.github.se.cyrcle.model.parking.PARKING_MAX_AREA
+import com.github.se.cyrcle.model.parking.TestInstancesParking
 import com.github.se.cyrcle.ui.map.MapConfig
 import com.mapbox.geojson.Point
 import com.mapbox.maps.CameraState
@@ -35,6 +38,9 @@ class MapViewModel : ViewModel() {
   private val _isAreaTooLarge: MutableStateFlow<Boolean> = MutableStateFlow(false)
   val isAreaTooLarge: StateFlow<Boolean> = _isAreaTooLarge
 
+  private val _userPosition = MutableStateFlow<Point>(TestInstancesParking.EPFLCenter)
+  val userPosition: StateFlow<Point> = _userPosition
+
   /**
    * Update the state of the location picker, This state is used to determine which steps of the
    * process to set the new location are completed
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
```diff
@@ -142,17 +157,31 @@ class MapViewModel : ViewModel() {
     return Pair(bottomLeftCorner, topRightCorner)
   }
 
+  private val handler = Handler(Looper.getMainLooper())
+  private var lastUpdateTime = 0L
+  private val updateInterval = 1000L // 1 second
```
Pijukai, line 162: For this value, there is a trade off between debouncing but making sure the app stays functionnal and display relevant infos concerning distances. Do you think 1 seconds is enough or maybe too much ? (Just a debate here). I personnally think 0.5 seconds could be better but that is just a minor detail.

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
```diff
@@ -181,6 +181,7 @@ class ParkingViewModel(
 
   /** set the center of the circle, and reset the radius to DEFAULT_RADIUS */
   fun setCircleCenter(center: Point) {
+    Log.d("ParkingViewModel", "Setting center to $center")
```
Pijukai, line 184: Forgot to delete it maybe or did you let it on purpose ?

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -114,7 +115,9 @@ fun MapScreen(
             .pitch(0.0)
             .zoom(maxZoom)
             .padding(EdgeInsets(100.0, 100.0, 100.0, 100.0))
-            .build())
+            .build()) {
+          mapViewportState.cameraState?.let { mapViewModel.updateUserPosition(it.center) }
```
Pijukai, line 119: don't forget to update the documentation a bit since you are updating the userPosition value in MapViewModel.

Pijukai:
By the way for the error you are getting in the CI right now, take a look at how you are managin the different context and activity. This was the solution on my part.

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
Armetiis: Forgot

app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
Armetiis: I will go with half a second. I don't think it matters so much in this screen. In my opinion, for the list screen we could even update once a minute / not update it. The user won't use it to refer themselves to a parking. Just to have a quick and easy overview

sean-prz (approved):
The feature works as intended on my phone with and without the location activated.
Nice job, I have nothing to add about the code.

Pijukai:
Small bug but when I click on list then go back to the map the radius on the list Screen quickly switch from 100m to 200m . do you have the same ?

Armetiis:
> Small bug but when I click on list then go back to the map the radius on the list Screen quickly switch from 100m to 200m . do you have the same ?

I am not sure this is related to my feature. Also it is not very clear what the bug is. Can you explain how to replicate it?

Pijukai:
Click on the ListScreen then you click on the map. During the transition, before the list screen disappear the 100m will go to 200m .

sean-prz:
I just noticed, the location is only updated on the map screen, never on the list screen I will see the distances update by itself even when I move. 
It's fine to merge the feature like this for now, I keep my approval

Pijukai:
That is fine by me the feature is functionnal and display the correct and expected behavior. This kind of improvement can be done later on.

Armetiis:
> I just noticed, the location is only updated on the map screen, never on the list screen will the user see the distances update by itself even when he moves. It's fine to merge the feature like this for now, I keep my approval

You're right. I think for now it's ok, as the user won't necessarily use the List Screen to guide themselves to a parking

Armetiis:
> Click on the ListScreen then you click on the map. During the transition, before the list screen disappear the 100m will go to 200m . This number refere to the one at the top of the List Screen telling the user which radius is used to filter the parkings.

I cannot replicate the bug. For me it starts at 1000m. Given that it's the max, it doesn't increase when I switch. I think we can leave this for another time. It's not a major issue

Pijukai (approved):
The feature works as expected and provide more relevant infos to the user Great job! The app starts to look more and more complete that's great to see.

Pijukai:
> > Click on the ListScreen then you click on the map. During the transition, before the list screen disappear the 100m will go to 200m . This number refere to the one at the top of the List Screen telling the user which radius is used to filter the parkings.
> 
> I cannot replicate the bug. For me it starts at 1000m. Given that it's the max, it doesn't increase when I switch. I think we can leave this for another time. It's not a major issue

Agreed the feature largely answers are need. that's a minor issue.

sean-prz (approved):

PR #199, opened 2024-11-13
Theme

### LINKED ISSUE #168: Use the Cyrcle logo for the app icon
#### Problem
Currently the icon of the app is the default android logo

#### Solution
Replace it with our logo

### LINKED ISSUE #198: Create a theme

andrepeiry:
# Finally a theme

After multiple tasks working on the theme, this PR should turn the page of the subject for a moment

## What

An overall lifting of the app, only done by using the theme and the atoms and molecules. Everything is now united though the triplet of color (from left to right) Cerulean, Shamrock Green and YInMn Blue Theme color triplet

The application logo was also redone due to bad manipulations in the last PR touching it. It will now appear during the launch of the application, the toast that are displayed, and will be used if our app ever use notification.
New application logo

## Can we see ? Can we see ?

Here's is a collection of the different screens that our app has so far

AddParkingAttribute_DropDown
AddParkingLocation
AddReview
AllReview_WithFilters
List_WithFilters
Map
ParkingDetails
Profile_Anonymous
Profile_SignedIn
SignIn

## What's next ?

Some possible ameliorations (but not limited to)

- Dark theme
- Colorblindness friendly design
- Customizable theme

Also consider the fact that this PR is mostly applying the theme onto the screen. Nothing was done in order to make the screens more coherent between them, nor UX improvements.

## Code coverage

Code coverage as of commit d4a87ae

## Closing issues

Close #168 
Close #198


sean-prz:
Great theme ! 
I just think we should not have any text in our primary. 
It's not like Easyjet had texts in orange or la poste had texts in yellow. 

I can send you a patch, I made this morning quickly to make the EnumDropdown molecules black except when focused as well as our InputTextfield.
It would look like this : 
image

The patch also changes the default text Color to hardcoded black. 

[fix-AddScrensTextToBlack.patch](https://github.com/user-attachments/files/17744329/fix-AddScrensTextToBlack.patch)

sean-prz:
Still to reduce the number of text in Primary you could change onSurface from our primary to the default black, so that the listScreen looks like this : 

image

app/src/main/java/com/github/se/cyrcle/ui/addParking/location/overlay/RectangleSelection.kt
```diff
@@ -74,15 +81,14 @@ fun RectangleSelection(
...
             drawRect(
-                color =
-                    if (isAreaTooLarge) Red.copy(alpha = 0.5f)
-                    else Color(0xFF22799B).copy(alpha = 0.7f),
+                color = if (isAreaTooLarge) Red.copy(alpha = 0.5f) else Cerulean.copy(alpha = 0.7f),
                 topLeft = canvasCenter,
                 size = rectSize)
 
             // Outline of the rectangle
+            drawRect(color = Cerulean.copy(alpha = 0.7f), topLeft = canvasCenter, size = rectSize)
```
Armetiis, line 89: Here you remove the if statement that displays the rectangle as red (with 50% opacity) if it is not valid

Armetiis (commented):
Really nice improvement. The theme looks beautiful and makes the UI of the app pop!! You did a great job
I however spotted some little problems

1. In the input text fields with discard icon, the icon is not visible
2. In the filters for a parking in the list screen, the checkbox is not visible
3. For the parking selection, I directly left you an in code comment

As for suggestions
I side with Sean on the fact that there's way too much text in primary color. We definitely need more black text. 
I however disagree about "We should not have ANY text in our primary". Primary can be nice for some small text in some occasions. (For example, in the screenshot he sent, having the "No reviews" text in primary adds charm to the UI).

Thank you for having taken the time to really make it look good

app/src/main/java/com/github/se/cyrcle/ui/addParking/location/overlay/RectangleSelection.kt
andrepeiry: A bad merge has occurred somewhere, I'll fix it ^^

andrepeiry:
# Resolving comments

At the moment this comment is posted, all things should have been resolve and pushed.

Please let me know if anything was forgotten.

## @sean-prz 

> (Make sure to not use onSurface anymore on the details screen, to keep the attributes standing out, next to their labels)

The goal of `onSurface` is to be used, I'm not sure I'm following you on this one. I've let the `ParkingDetailsScreen` as it was, because having everything in black in simply unreadable (because everything is the same color, distinguishing label from content is more difficult, even with the padding).

> [fix-AddScrensTextToBlack.patch](https://github.com/user-attachments/files/17744329/fix-AddScrensTextToBlack.patch)
> [fix-listScreenTextToBlack.patch](https://github.com/user-attachments/files/17744512/fix-listScreenTextToBlack.patch)

Thanks but no thanks, the colors are being written in multiple places, and I spend 3 tasks doing the exact opposite. I've reworked them so that it better fit in the codebase.

## @Armetiis 

> Also another kind remark (nonblocking), is the outline of the toggle switch for the advanced mode. I don't think it looks really good in our primary

That is more a design decision than a theme one, I'll let you discuss this with the creator/designer of this button

> I however disagree about "We should not have ANY text in our primary".

Given the intensive lobbyism done by Sean on this, I've gone with his solution. But this can be discussed in a design meeting sometimes in the future, as this is more a collective decision

sean-prz (commented):
Thanks for the changes
I am now in love with our UI. 

However, the radio button on the list screen to filter the camera surveillance is still hidden.

sean-prz (approved):
Looks like we are ready for M2

PR #204, opened 2024-11-14
Coin system skeleton

### LINKED ISSUE #200: Coin System Backend

andrepeiry:
# What ?

A basis for a future coin and currency system in our app !

## How

Add a Wallet property to the user, as well as a dedicated view model for ease of use.

## Closing issues

Close #200 

## Code Coverage

Code Coverage as of 02ba85c6

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/model/user/Wallet.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/user/Wallet.kt
@@ -1,0 +1,24 @@
+package com.github.se.cyrcle.model.user
+
+typealias Coin = Int
+
+/**
+ * TODO
+ */
+class Wallet {
+
+  private var coins: Coin = 0
+
+  fun isSolvable(coinsToHave: Coin): Boolean {
+    return coins >= coinsToHave
+  }
+
+  /**
+   * TODO
+   */
+    companion object {
+        fun empty(): Wallet {
+            return Wallet()
+        }
+    }
+}
--- a/app/src/main/java/com/github/se/cyrcle/model/user/Wallet.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/user/Wallet.kt
@@ -16,6 +16,7 @@ class Wallet(private var coins: Coin = 0) {
    * Add amount of coins to the wallet.
    *
    * @param amount the amount of coins to add. Must be positive.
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/se/cyrcle/model/user/Wallet.kt
```diff
@@ -0,0 +1,59 @@
...
+  /**
+   * Return the amount of coins currently in the mallet. WARNING: This method does not modify the
+   * value of the coins in the wallet.
+   */
+  fun getCoins(): Coin {
+    return coins
+  }
+
+  /** Creates an empty [Wallet]. It contains no coins. */
+  companion object {
```
matttsss, line 54: I'm not sure if the `companion object` is realy necesary, we already have a default argument for that?

matttsss (commented):
A pretty good start for our community encouragment system. I just have a few doubts about why we would need a anonymous coin system? As you said it is not persistent and the anonymous user can't contribute.

matttsss:
Also, most of the exceptions aren't documented... Could you maybe do something for that?

app/src/main/java/com/github/se/cyrcle/model/user/Wallet.kt
andrepeiry: It is maybe a bit overkill, but as the feature isn't well define in the project, I preferred doing a bit too much than too little (image if, by default, users have a welcome gift, then `Wallet(coin = 0)` make less sens.

matttsss: Right... that makes sense!

AlexandreBeeDeeBop (changes requested):
Are you sure the constant for the threshold should be in the CoinViewModel ? I would think that a better place would be at the instantiation, so that we can better modularize the code.

andrepeiry:
> Are you sure the constant for the threshold should be in the CoinViewModel ? I would think that a better place would be at the instantiation, so that we can better modularize the code.

That's a good thought, I'll update the class to be more modular ^^

AlexandreBeeDeeBop (approved):
This is getting exciting ! Soon we'll be able to gamble ! 
Jokes aside, nothing else to say, it's all good.

matttsss (approved):
All good for me, you are free to roll 👌

PR #244, opened 2024-11-18
Feature add parking details navigation

AlexandreBeeDeeBop:
Content of the PR:
This trivial PR simply adds the possibility to navigate from the favorite parking card in the user's profile screen to its detail's screen. The favorite parking cards are now clickable and send you to the selected parking's detail screen. There is also a small test added to check this behaviour. Nothing much but still an essential feature.

Future ideas : 
I think the favorite parking cards need more than just the names on it.
In a future sprint, I would like to add either a few useful information about the parking, or the possibility to add personal notes beneath the names.


LouisJamesVasseur (approved):
Cool feature! It works as advertised and code is good too!

andrepeiry (approved):
Code is pretty straight-forward

I would just note that the `val context` at `ViewProfileScreen:221` is used only once. If it is needed as a value, please declare it as near as possible to its usage.

PR #245, opened 2024-11-18
Complete the image repository 

sean-prz:
_closes #246_
## This PR reworks the image repository
The image repository had a first version implemented but it was never used.
This version is ready to be used for the view model, we now have a full working upload/download mechanism in the repository

### Key Changes 
 - **implement the upload** _see more details in the How section_

### Minor Changes
- move the repository into its own package (parking->image)
- use injection for constructor parameter (defined the necessary in AppModule)
- replace suspend by callback (to stay in line with the rest of the project)
- implement the upload 
- adjust the mock
- add unit test

## How does this repo. works. 

The function `getUrl` is used when wanting to download an image.
The path to the file on the cloud storage is stored in a **firestore** object. 
This function translate it into an URL that can be passed to the UI layer. 

The function `uploadImage` takes a local path file and upload a compressed version of its content to our cloud. The viewmodel will have the responsability to store into a firestore object the path. 
 

## Why this PR
This prepare the backend with the goal of being able to upload and download picture from our app, both for the user profile picture and for the parking images

image

### DIFF
--- a/app/src/androidTest/java/com/github/se/cyrcle/ui/addParking/AddScreensNavigationTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/ui/addParking/AddScreensNavigationTest.kt
@@ -32,6 +32,7 @@ import com.github.se.cyrcle.ui.addParking.location.LocationPicker
 import com.github.se.cyrcle.ui.map.MapScreen
 import com.github.se.cyrcle.ui.navigation.NavigationActions
 import com.mapbox.geojson.Point
+import dagger.hilt.android.testing.HiltAndroidTest
 import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
@@ -39,6 +40,7 @@ import org.mockito.Mock
 import org.mockito.Mockito.`when`
 import org.mockito.MockitoAnnotations
 
+@HiltAndroidTest
 @RunWith(AndroidJUnit4::class)
 class AddScreensNavigationTest {
   @get:Rule val composeTestRule = createComposeRule()


Armetiis (approved):
Can't wait to actually upload my profile picture to the database! Good job

app/src/test/java/com/github/se/cyrcle/model/image/ImageRepositoryCloudStorageTest.kt
```diff
@@ -0,0 +1,152 @@
...
+    val taskCompletionSource = TaskCompletionSource<Uri>()
+    val mockUploadTask = mock(UploadTask::class.java)
+    `when`(mockAuth.currentUser).thenReturn(mock())
+    `when`(mockContext.contentResolver).thenReturn(mockContentResolver)
+    `when`(mockContentResolver.openInputStream(Uri.parse(fileUri)))
+        .thenReturn(ByteArrayInputStream(bytearray))
+    `when`(mockStorageReference.child(destinationPath)).thenReturn(mockStorageReference)
+    `when`(mockStorageReference.putBytes(any())).thenReturn(mockUploadTask)
+    `when`(mockStorageReference.downloadUrl).thenReturn(taskCompletionSource.task)
+    `when`(mockUploadTask.addOnSuccessListener(any())).thenAnswer { invocation ->
```
matttsss, line 106: I admire your motivation to mock...

matttsss (commented):
So if I get it correctly, for the moment there is no UI or ViewModel use of this right?
What are the plans to use the repository?

app/src/test/java/com/github/se/cyrcle/model/image/ImageRepositoryCloudStorageTest.kt
sean-prz: I don't think I had the choice ... did I ?

sean-prz:
@matttsss 
Indeed there is no way for you to check this is working, as there is no UI/viewmodel implementation.

I did check that this implementation is working on a branch where the UI/viewmodel/repo is integrated and working but I can't push it all in one PR it would be huge.

- The viewmodel who chooses the path on the cloud storage 
- The viewmodel store this path into the firestore document of the object (either the user or the parking)
- The viewmodel then pass this cloud storage path as well as the path of the picture on the user device to the repo. 

I'll make another wiki section that explains how all of this works just like I did for the fetch of the parkings.

app/src/test/java/com/github/se/cyrcle/model/image/ImageRepositoryCloudStorageTest.kt
matttsss: Not really :)

matttsss (approved):
All right! I was afraid that the path might be dependent on the repository but since it's only relative, I am all good with this!

sean-prz:
> All right! I was afraid that the path might be dependent on the repository but since it's only relative, I am all good with this!

I did not consider this point, but my implementation respects it.
This is another reason that I never thought about of why we can't just store URL in the firestore document, the url depends on the implementation while the path doesn't, neat.

PR #249, opened 2024-11-19
Prepare the user data class for pictures

sean-prz:
_closes #250_

## What
This PR prepares the user data so that we will be able to download and upload picture :

## How
You can find a graph below that summarize the new user data class, but I'll quickly describe the changes: 

- The public URL has been modified to be a  public cloud path (i.e "ProfilePictures/axwez.jpg")
- The URL is not stored in the public part of the user but in a new part called session
- The session part of the user Data class holds informations that don't need to be backed-up.
- The session part also hold the URI (the local path on the user device) 
Used when he selected the picture to pass it to the backend and also to preview the picture before sending it to the cloud.

_A graph of the structure of the user data class_
image

## Why 
One field to hold the 3 objects  (Uri, Url, cloud path) was not enough, since they are clearly different even if they are used to display pictures, this is easier to maintain and understand 

-- -
image

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/model/user/User.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/user/User.kt
@@ -5,13 +5,13 @@ package com.github.se.cyrcle.model.user
  *
  * @property userId The unique identifier of the user.
  * @property username The username of the user.
- * @property profilePictureUrl The path to the profile picture of the user on the cloud storage.
- *   (Not implemented)
+ * @property profilePictureCloudPath The path to the profile picture of the user on the cloud
+ *   storage. (Not implemented)
  */
 data class UserPublic(
     val userId: String,
     val username: String,
-    val profilePictureUrl: String = "",
+    val profilePictureCloudPath: String = "",
     // val accountCreationDate: Timestamp? = null,
     // val numberOfContributedSpots: Int = 0,
     // val userReputationScore: Int = 0
@@ -36,9 +36,26 @@ data class UserDetails(
 )
 
 /**
+ * For information that are only relevant to the session and won't be stored in the database
+ *
+ * @property profilePictureUrl The URL of the profile picture
+ * @property profilePictureUri The URI of the profile picture,(local path on the device)
+ */
+data class LocalSession(
+    val profilePictureUrl: String? = "",
+    val profilePictureUri: String? = "",
+)
+
+/**
  * Data class representing a user.
  *
  * @property public Public information of the user.
  * @property details Private information of the user.
+ * @property localSession Information that are only relevant to the session and won't be stored in
+ *   the database
  */
-data class User(val public: UserPublic, val details: UserDetails?)
+data class User(
+    val public: UserPublic,
+    val details: UserDetails?,
+    val localSession: LocalSession? = null
+)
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/github/se/cyrcle/ui/profile/EditProfileComponent.kt
```diff
@@ -100,9 +99,11 @@ fun EditProfileComponent(
...
 
           saveButton(
-              User(
-                  UserPublic(user.public.userId, username, profilePictureUrl),
-                  UserDetails(firstName, lastName, user.details!!.email)))
+              // Use .copy to not lose the locas session and avoid re-fetching the user from db.
+              user.copy(
+                  public = user.public.copy(username = username, profilePictureCloudPath = ""),
+                  details = user.details?.copy(firstName = firstName, lastName = lastName),
+                  localSession = LocalSession(profilePictureUri = profilePictureUri)))
```
andrepeiry, line 106: One would argue that argument with the same name as the variable given would not be necessary. Nothing blocking, just that it makes code heavier to read for no reason.

app/src/main/java/com/github/se/cyrcle/model/user/User.kt
```diff
@@ -5,13 +5,13 @@ package com.github.se.cyrcle.model.user
...
- * @property profilePictureUrl The path to the profile picture of the user on the cloud storage.
- *   (Not implemented)
+ * @property profilePictureCloudPath The path to the profile picture of the user on the cloud
+ *   storage. (Not implemented)
  */
 data class UserPublic(
     val userId: String,
     val username: String,
-    val profilePictureUrl: String = "",
+    val profilePictureCloudPath: String = "",
```
andrepeiry, line 14: Non-blocking, but I'm not a big fan of the term "Cloud". I have nothing in mind, maybe `ServerStorage` or something, but I'm not even sure it can be describe shortly otherwise.

app/src/main/java/com/github/se/cyrcle/model/user/User.kt
```diff
@@ -5,13 +5,13 @@ package com.github.se.cyrcle.model.user
  *
  * @property userId The unique identifier of the user.
  * @property username The username of the user.
- * @property profilePictureUrl The path to the profile picture of the user on the cloud storage.
- *   (Not implemented)
+ * @property profilePictureCloudPath The path to the profile picture of the user on the cloud
+ *   storage. (Not implemented)
```
andrepeiry, line 9: Even though it is not this PR, having `Not implemented` in the documentation would be bad practice. Having dependency in the documentation (because that's what we're gonna have to do once Cloud storage is up, will be to update said documentation. If a system is not yet implemented, only this system should be marked as such. If I'm mistaken and this classes are said system, then it would be better to have it written in big inside the description, not the property/param.

app/src/main/java/com/github/se/cyrcle/model/user/User.kt
sean-prz: yeah I was trying to find a name that doesn't depend on the current implementation (that is cloud storage) but also that make it clear that it is not the path of the typical firestore storage  we use in the rest of the project

app/src/main/java/com/github/se/cyrcle/ui/profile/EditProfileComponent.kt
sean-prz: I'll remove it !

app/src/main/java/com/github/se/cyrcle/model/user/User.kt
andrepeiry: I think that sometimes, naming variables really is one of the hardest things in programming ^^

PR #252, opened 2024-11-20
Profile pictures are persistent

sean-prz:
_closes #253 #254_
**Important note** : Delete your user document on firestore before testing this PR

## What
The user's profile picture is send to a database, making it persistent, and retrieve on log-in to be displayed.

### How
_The viewmodel had to be changed, to send request to the image repository, the changes are quite complex, so the easiest way to understand what is happening is to look at the graphs I prepared for the wiki_  : 

On sign-in : 
image

On modify profile : 
image

### Major changes
- A new function in the viewmodel `uploadPicture` is called only when `updatesUser` detects an URI has been set in the updatedUser
- A new function `transferPathToUrl` is called when we want to fetch the profile picture of a user, or our own profile pictures on login, it is called inside `getUserById(user, onSuccess)`

### Minor changes :
- The function of the viewmodel `getUserById(user: User)` that actually set the current user has been renamed to `setCurrentUserById() `
- `AddUser` has been renamed into `signIn` to better reflect what it does .
-  A SignOut function has been introduced instead of calling `setCurrentUser(null)` (In preperation for making `setCurrentUser(null)` null
-  The `userviewmodel` can be constructed without an `ImageRepository` and won't handle picture interactions this is for backward compability with the test. 

PS: I left the debugging log because they are really useful here to understand what happens and will probably be useful in the future. 
-- -
image

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
@@ -126,7 +126,7 @@ class UserViewModel(
     // First update the user object in the view model.
     setCurrentUser(user)
     // Then update the user object in the database
-    if (user.localSession?.profilePictureUri == null || context == null) {
+    if (user.localSession?.profilePictureUri.isNullOrBlank() || context == null) {
       userRepository.updateUser(
           user,
           onSuccess = { Log.d("UserViewModel", "User updated in the database") },
--- a/app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
@@ -20,7 +20,7 @@ import kotlinx.coroutines.flow.map
 class UserViewModel(
     private val userRepository: UserRepository,
     private val parkingRepository: ParkingRepository,
-    private val imageRepository: ImageRepository
+    private val imageRepository: ImageRepository? = null
 ) : ViewModel() {
 
   private val _currentUser = MutableStateFlow<User?>(null)
@@ -34,7 +34,7 @@ class UserViewModel(
    *
    * @param user the user to set as the current user
    */
-  private fun setCurrentUser(user: User?) {
+  fun setCurrentUser(user: User?) {
     Log.d("UserViewModel", "Setting the current user in viewmodel : $user")
     _currentUser.value = user
   }
@@ -98,7 +98,7 @@ class UserViewModel(
   fun signIn(user: User) {
     // Set the user in the viewmodel even if it's incomplete as we need the ID to match to call
     // updateUser()
-    setCurrentUser(user)
+    // setCurrentUser(user)
     userRepository.addUser(
         user,
         onSuccess = { getUserById(user.public.userId) { setCurrentUser(it) } },
@@ -190,6 +190,10 @@ class UserViewModel(
    * path
    */
   private fun uploadProfilePicture(context: Context) {
+    if (imageRepository == null) {
+      Log.e("UserViewModel", "ImageRepository is null")
+      return
+    }
     val user = currentUser.value ?: return
     val fileUri = user.localSession?.profilePictureUri ?: return
     val destinationPath = "profile_pictures/${user.public.userId}"
@@ -216,6 +220,10 @@ class UserViewModel(
    * @param onSuccess the callback to call with the updated user
    */
   private fun transformPathToUrl(user: User, onSuccess: (User) -> Unit, onFailure: () -> Unit) {
+    if (imageRepository == null) {
+      Log.e("UserViewModel", "ImageRepository is null")
+      return
+    }
     val cloudPath = user.public.profilePictureCloudPath
     if (cloudPath.isBlank()) {
       Log.d("UserViewModel", "No image to fetch for user")
--- a/app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
@@ -1,13 +1,17 @@
 package com.github.se.cyrcle.model.user
 
+import android.content.Context
 import android.util.Log
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.ViewModelProvider
+import com.github.se.cyrcle.model.image.ImageRepository
+import com.github.se.cyrcle.model.image.ImageRepositoryCloudStorage
 import com.github.se.cyrcle.model.parking.Parking
 import com.github.se.cyrcle.model.parking.ParkingRepository
 import com.github.se.cyrcle.model.parking.ParkingRepositoryFirestore
 import com.google.firebase.auth.FirebaseAuth
 import com.google.firebase.firestore.FirebaseFirestore
+import com.google.firebase.storage.FirebaseStorage
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
@@ -15,7 +19,8 @@ import kotlinx.coroutines.flow.map
 
 class UserViewModel(
     private val userRepository: UserRepository,
-    private val parkingRepository: ParkingRepository
+    private val parkingRepository: ParkingRepository,
+    private val imageRepository: ImageRepository
 ) : ViewModel() {
 
   private val _currentUser = MutableStateFlow<User?>(null)
@@ -29,19 +34,25 @@ class UserViewModel(
    *
    * @param user the user to set as the current user
    */
-  fun setCurrentUser(user: User?) {
+  private fun setCurrentUser(user: User?) {
+    Log.d("UserViewModel", "Setting the current user in viewmodel : $user")
     _currentUser.value = user
   }
 
   private val _favoriteParkings = MutableStateFlow<List<Parking>>(emptyList())
   val favoriteParkings: StateFlow<List<Parking>> = _favoriteParkings
 
+  /** Signs out the current user by setting the state to null */
+  fun signOut() {
+    setCurrentUser(null)
+  }
+
   /**
-   * Gets a user by ID from the Firestore database and sets it as the current user.
+   * Set the current user by fetching the user from the Firestore database with the given ID.
    *
    * @param userId the ID of the user to get
    */
-  fun getUserById(userId: String) {
+  fun setCurrentUserById(userId: String) {
     userRepository.getUserById(
         userId,
         onSuccess = { setCurrentUser(it) },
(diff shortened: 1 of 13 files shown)


Pijukai:
Thanks for the graphs they are really helpful :1st_place_medal: . Since you will add them to the wiki be careful you wrote User porfile in the on sign in graph

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
```diff
@@ -153,6 +184,68 @@ class UserViewModel(
...
+        onFailure = { Log.e("UserViewModel", "Failed to upload image") })
+  }
+
+  /**
+   * Updates the user object with the profile picture url fetched from the repo.
+   *
+   * @param user the user object to update (doesn't have to be the current user)
+   * @param onSuccess the callback to call with the updated user
+   */
+  private fun transformPathToUrl(user: User, onSuccess: (User) -> Unit, onFailure: () -> Unit) {
```
Pijukai, line 222: you forgot to document the `onFailure` here

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
```diff
@@ -95,20 +111,35 @@ class UserViewModel(
...
-    userRepository.updateUser(
-        user,
-        onSuccess = { Log.d("UserViewModel", "User updated") },
-        onFailure = { exception ->
-          Log.e(
-              "com.github.se.cyrcle.model.user.UserViewModel",
-              "Failed to update user: ${user.public.userId}",
-              exception)
-        })
+  fun updateUser(user: User, context: Context? = null) {
```
Pijukai, line 119: You forgot to document the `context` argument here as well.

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
```diff
@@ -95,20 +111,35 @@ class UserViewModel(
...
+    }
+
+    Log.d("UserViewModel", "Updating user: $user")
+    // First update the user object in the view model.
+    setCurrentUser(user)
+    // Then update the user object in the database
+    if (user.localSession?.profilePictureUri == null || context == null) {
+      userRepository.updateUser(
+          user,
+          onSuccess = { Log.d("UserViewModel", "User updated in the database") },
```
Pijukai, line 132: For such a central function i would add an argument `OnSuccess` and put your log as default parameter so that the function stays as modular and flexible as it could be what do you think ?

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
```diff
@@ -95,20 +111,35 @@ class UserViewModel(
...
+          onFailure = { exception ->
+            Log.e(
+                "com.github.se.cyrcle.model.user.UserViewModel",
+                "Failed to update user: ${user.public.userId}",
+                exception)
+          })
+    } else {
+      Log.d("UserViewModel", "User has a profile picture to upload")
+      uploadProfilePicture(context)
```
Pijukai, line 141: What happens if the user modify his picture but at the same time change his username ? Shouldn't you update the user anyway ?

LouisJamesVasseur (changes requested):
Hi, firstly the feature itself works, good job!
Unfortunately it introduced a rather major bug alongside it where clicking on Save after modifying the account once a profile picture has been added crashes the app. I suspect that a few null checks or paths are missing in the UserViewModel, as @Pijukai has brought up.

sean-prz:
@LouisJamesVasseur The bug should be fixed : 

The viewmodel was checking if there was a picture to upload and wrongfully thought there was a picture to upload when there wasn't any. This made the app crash.

The viewmodel checked an attribute of the user, that is the profilePictureUri but the condition was wrong, I replaced `== null ` by `.isNullOrBlank`

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
sean-prz: Ah, I made this confusing sorry, the `uploadProfilePicture` body calls `updateUser` once the profile picture has been uploaded. 

Do you think adding a comment on line 140 would make it less confusing ?

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
sean-prz: Sure, I can add one, I don't see an immediate reason for it but it doesn't hurt to but one and make the code more flexible

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
sean-prz: Solved

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
sean-prz: Solved

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
LouisJamesVasseur: I think so (I mean, in general if there's a risk it may lead to confusion, better safe than sorry w/ comments :)) )

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
```diff
@@ -153,6 +192,76 @@ class UserViewModel(
...
+    val fileUri = user.localSession?.profilePictureUri ?: return
+    val destinationPath = "profile_pictures/${user.public.userId}"
+    imageRepository.uploadImage(
+        context = context,
+        fileUri = fileUri,
+        destinationPath = destinationPath,
+        onSuccess = { url ->
+          Log.d("UserViewModel", "Image uploaded successfully to $url")
+          val updatedUser =
+              user.copy(
```
LouisJamesVasseur, line 217: why do you copy the User, then use the update function? (I know it's functionally identical to addUser, but the way you've implemented it as is a little confusing...)

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
```diff
@@ -153,6 +184,68 @@ class UserViewModel(
...
+   * Upload the picture to the cloud storage update the user object in the database with the cloud
+   * path created. update the user object in the viewmodel with the url of the image and the cloud
+   * path
+   */
+  private fun uploadProfilePicture(context: Context) {
+    if (imageRepository == null) {
+      Log.e("UserViewModel", "ImageRepository is null")
+      return
+    }
+    val user = currentUser.value ?: return
```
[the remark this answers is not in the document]
sean-prz: This function uploadPicture is never supposed to be called when a user is not set.

I can only see this happens in test if we made a mistake writing the code.

I could go harder and instead of just returning throwing a fatal exception what do you think ?

app/src/main/java/com/github/se/cyrcle/ui/profile/EditProfileComponent.kt
```diff
@@ -48,17 +48,21 @@ fun EditProfileComponent(
...
   var lastName by remember { mutableStateOf(user.details!!.lastName) }
   var profilePictureUri by remember { mutableStateOf(user.localSession?.profilePictureUri) }
+  val profilePictureUrl by remember { mutableStateOf(user.localSession?.profilePictureUrl) }
+  // To display the local profile picture when the user changes it, and hasn't uploaded (saved) it
+  var hasTemporarilyChangedProfilePicture by remember { mutableStateOf(false) }
 
   val imagePickerLauncher =
       rememberLauncherForActivityResult(ActivityResultContracts.GetContent()) { uri ->
         uri?.let { profilePictureUri = it.toString() }
+        hasTemporarilyChangedProfilePicture = true
```
LouisJamesVasseur, line 58: what's the use of the mutableState here? it starts at false then is statically moved to true, so I don't see its utility, probably just me though...

LouisJamesVasseur (approved):
Thanks for the fix! Everything looks good now, I just added a couple non-blocking comments for you to consider. Great work!

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
Pijukai: The comment would be useful for us since we are going deeper in the code if it is okay for you could be a good idea.

app/src/main/java/com/github/se/cyrcle/ui/profile/EditProfileComponent.kt
sean-prz: I added a comment where I declare it to make it clear, but if it is not clear enough :


When you have selected the new profile picture but haven't click save yet, you want to see your profile with the new picture you just selected, but the picture hasn't been uploaded yet so you have to get it from the local storage this is where this state comes in to switch between displaying local storage or from the database

LouisJamesVasseur: Thanks for clarifying!

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
sean-prz: Not sure I understand

User is immutable if I want to modify some fields I need to do a .copy, here I want to update the user with the new Url I just got, so I do a copy then I call updateUser so that it replaces the user document and the user in the viewmodel with this new instance

This is the same mechanism as in the edit profile screen.

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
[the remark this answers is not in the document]
Pijukai: Well if you are sure that this function should never be called when currentUser is null then you can. However i was thinking more of a Log which seems less "fatal" xD.

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
sean-prz: Solved

Pijukai (approved):
Great job your code is well written and documented. Thank you for your quick fixes. APPROVED !

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
sean-prz: I added some explanation in the latest commit

Pijukai (approved):
Well done ! Reaproving because my precedent approval was stale

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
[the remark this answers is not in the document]
sean-prz: Solved, added Log.e() + return if invalid arguments

PR #255, opened 2024-11-20
Feature CreateProfileScreen

### LINKED ISSUE #110: Make a user "Account Creation" screen if the user doesn't already exist
We need to implement a new screen for user account creation when a new user signs in for the first time.

### Current Situation
Users can sign in using their Google accounts.
There's no dedicated screen for new users to complete their account setup.

### Desired Outcome
When a new user signs in for the first time, they should be directed to an "Account Creation" screen.
This screen should allow users to:
Review and confirm their basic information (e.g., name, email) pulled from their Google account.

### Potential future steps
Make user accept terms of service and privacy policy.

matttsss:
## What is this PR?
This PR introduces a new screen allowing users to create their account.

## How?
A new button was added to the `SignInScreen`, it brings users to a new screen that allows them to create their account.

### Pitfals:
Currently, no checks are made to check the presence of an already existing user. 
Also, the button for creating the account can be a bit confusing. There is no control over it's appearance since it is issued by the `Authenticator`. This will have to be change in a refactor for the `Authenticator` system.
Both of these points could be worked on as a part of a later sprint.

## Features:
image
image

### Test coverage:
image

CLoses #110

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/profile/EditProfileComponent.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/profile/EditProfileComponent.kt
@@ -20,8 +20,9 @@ import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
 import com.github.se.cyrcle.R
-import com.github.se.cyrcle.model.user.LocalSession
 import com.github.se.cyrcle.model.user.User
+import com.github.se.cyrcle.model.user.UserDetails
+import com.github.se.cyrcle.model.user.UserPublic
 import com.github.se.cyrcle.ui.theme.atoms.ConditionCheckingInputText
 import com.github.se.cyrcle.ui.theme.atoms.Text
 
@@ -112,10 +113,12 @@ fun EditProfileComponent(
           saveButton(copyUser(user))
           cancelButton()
         } else {
-          Row(horizontalArrangement = Arrangement.spacedBy(8.dp)) {
-            cancelButton()
-            saveButton(copyUser(user))
-          }
+          Row(
+              modifier = Modifier.testTag("EditComponentButtonRow"),
+              horizontalArrangement = Arrangement.spacedBy(8.dp)) {
+                cancelButton()
+                saveButton(copyUser(user))
+              }
         }
       }
 }
--- a/app/src/main/java/com/github/se/cyrcle/ui/profile/EditProfileComponent.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/profile/EditProfileComponent.kt
@@ -39,6 +39,7 @@ fun EditProfileComponent(
     user: User?,
     saveButton: @Composable (User) -> Unit,
     cancelButton: @Composable () -> Unit,
+    modifier: Modifier = Modifier,
     verticalButtonDisplay: Boolean = false
 ) {
   if (user == null) {
@@ -69,7 +70,7 @@ fun EditProfileComponent(
       }
 
   Column(
-      modifier = Modifier.fillMaxSize().padding(16.dp).testTag("ProfileContent"),
+      modifier = modifier.fillMaxSize().padding(top = 15.dp).testTag("ProfileContent"),
       horizontalAlignment = Alignment.CenterHorizontally) {
         ProfileImageComponent(
             url = if (hasTemporarilyChangedProfilePicture) profilePictureUri else profilePictureUrl,
--- a/app/src/main/java/com/github/se/cyrcle/ui/profile/EditProfileComponent.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/profile/EditProfileComponent.kt
@@ -20,9 +20,8 @@ import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
 import com.github.se.cyrcle.R
+import com.github.se.cyrcle.model.user.LocalSession
 import com.github.se.cyrcle.model.user.User
-import com.github.se.cyrcle.model.user.UserDetails
-import com.github.se.cyrcle.model.user.UserPublic
 import com.github.se.cyrcle.ui.theme.atoms.ConditionCheckingInputText
 import com.github.se.cyrcle.ui.theme.atoms.Text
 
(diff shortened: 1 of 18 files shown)


Pijukai:
I just looked at your pitfalls. I agree that the button is confusing but can be corrected during later sprints. However is there no way of checking already existing users ? Because having the different accounts linked to the same google account seems like a big pitfall that should be addressed during this PR/Sprint. Is it possible to maybe add a quick security measure ?

app/src/main/java/com/github/se/cyrcle/ui/profile/InviteToAuthScreen.kt
```diff
@@ -0,0 +1,58 @@
...
+import androidx.compose.ui.unit.dp
+import com.github.se.cyrcle.R
+import com.github.se.cyrcle.ui.navigation.NavigationActions
+import com.github.se.cyrcle.ui.navigation.Route
+import com.github.se.cyrcle.ui.theme.atoms.Button
+import com.github.se.cyrcle.ui.theme.atoms.Text
+import com.github.se.cyrcle.ui.theme.molecules.BottomNavigationBar
+
+@Composable
+fun InviteToAuthScreen(navigationActions: NavigationActions) {
```
Pijukai, line 26: Could you maybe add some description about this screen in the PR description. I am a bit confused as to what it is.

app/src/main/java/com/github/se/cyrcle/ui/profile/InviteToAuthScreen.kt
```diff
@@ -0,0 +1,58 @@
...
+            stringResource(R.string.invite_auth_screen_invite_to_sign_in),
+            style = MaterialTheme.typography.titleLarge)
+
+        Spacer(modifier = Modifier.height(16.dp))
+
+        Text(stringResource(R.string.invite_auth_screen_give_reasons_to_sign_in))
+
+        Spacer(modifier = Modifier.height(8.dp))
+
+        Button(
```
Pijukai, line 47: I think testing the display of the button is also relevant added to the test of the Scaffold what do you think ?

app/src/main/java/com/github/se/cyrcle/ui/profile/EditProfileComponent.kt
```diff
@@ -52,14 +54,23 @@ fun EditProfileComponent(
   // To display the local profile picture when the user changes it, and hasn't uploaded (saved) it
   var hasTemporarilyChangedProfilePicture by remember { mutableStateOf(false) }
 
+  fun copyUser(user: User): User {
```
Pijukai, line 57: Don't forget the documentation here

app/src/main/java/com/github/se/cyrcle/ui/profile/EditProfileComponent.kt
```diff
@@ -52,14 +54,23 @@ fun EditProfileComponent(
...
+
   val imagePickerLauncher =
       rememberLauncherForActivityResult(ActivityResultContracts.GetContent()) { uri ->
         uri?.let { profilePictureUri = it.toString() }
         hasTemporarilyChangedProfilePicture = true
       }
 
   Column(
-      modifier = Modifier.fillMaxSize().padding(16.dp).testTag("ProfileContent"),
+      modifier = modifier.fillMaxSize().padding(top = 15.dp).testTag("ProfileContent"),
```
Pijukai, line 73: I don't think this test tag is used in the test. Is it intended ?

sean-prz:
image
the images dos not t display well on the PR

sean-prz:
Tell me if I understood right, you want to make, in another PR, these two checks that are not yet implemented : 
- A user without an account should not be able to login
- A user with an account should not be able to create an account.

To me it's crucial that these checks are made, but If you plan on doing it later, no problem we can merge a first version like this without the check.
 I just want to make sure we agree on the check we will have to implement

app/src/main/java/com/github/se/cyrcle/ui/profile/EditProfileComponent.kt
matttsss: There are so many ways to copy users... I just made a inner function because I used it twice and it could be a tricky bug if both copies are different. I don't think it should be give much more importance than that...

app/src/main/java/com/github/se/cyrcle/ui/profile/EditProfileComponent.kt
matttsss: Well spoted... But I don't think it can hurt to keep it, it might be usefull in later tests (E2E for example)

matttsss:
@sean-prz @Pijukai I can understand you opinions on the checks. For the moment, without the check, the CreateProfileScreen just works as a EditProfileScreen if the User already exists. I did not find it bothersome since there are no Milestones between now and next week and except for our team there wouldn't be anyone else using our app. I do not have time to add a check for this week. What I could suggest is to comment out the button that leads to CreateProfileScreen and make this PR a "silent feature" if you realy do not like it.
(and yes Sean those are the planned checks I would add next week)

Pijukai:
@matts I was a bit confused about it but after reviewing the code, it is true that it is not bothersome. I just misunderstood how the logic worked but now everything imake sense and it does not sound as dramatic as sean presented it :laughing:.

app/src/main/java/com/github/se/cyrcle/ui/profile/EditProfileComponent.kt
Pijukai: I agree with you on this point , i was just checking that you didn't missed it during your tests even though you planned on using it

app/src/main/java/com/github/se/cyrcle/ui/profile/EditProfileComponent.kt
Pijukai: It was just a question of organization, i am used to writing helping function outside of every other function if possible but that worked as well.

Matts:
@Pijukai wrong @matttsss, have a good evening.

Pijukai:
> @Pijukai wrong @matttsss, have a good evening.

Oups sorry my bad 2 similar username , have a good evening.

Pijukai:
@matttsss when i test the create profile screen the username text field is already red even though nothing was typed in, is it possible to make something for it because it seems unusual for me at least.

sean-prz:
I think you already knew that but in case : 
In the following scenario :
- I already have an account
- I click create an account 
- I modify fields save and log in
Then because of the way sign-in works the changes the user made won't be send to the repo and saved :
image

This is a good things since we don't want a user who forgot that he already had an account to overwrite his old profile.
-- -
I believe the best things to do in the future would be : 

on the createProfile : 
image

on the SignInScreen
image

sean-prz (approved):
Features seems to work !
Thanks for the screenshots in the UI too

Looking forward to force new user to go through this screen !

matttsss:
> @matttsss when i test the create profile screen the username text field is already red even though nothing was typed in, is it possible to make something for it because it seems unusual for me at least.

This is out of my control, it is in one of the molecules in the repo. :sweat_smile:

Pijukai (approved):
Great feature well done. I didn't really understand how the logic worked but thnaks to you everything is clear. APPROVED !

Pijukai:
> > @matttsss when i test the create profile screen the username text field is already red even though nothing was typed in, is it possible to make something for it because it seems unusual for me at least.
> 
> This is out of my control, it is in one of the molecules in the repo. 😅

Nevermind then it's not that important

PR #256, opened 2024-11-20
bugfix : non persistent state in view model during Configuration changes

### LINKED ISSUE #239: BugFix: Switching light mode to dark mode reset part of the state of the app
The problem is that the activity/viewModels got destroyed after the configuration changes (switching to dark mode). This creates unexpected behaviors such as forced sign out when switching mode.

Pijukai:
## PR content

This PR fixes the bug linked to the state and view models being destroyed when the configuration changes (e.g switch to dark theme)

## Fix

The fix was to refactor the view models outside the `OnCreate` of the MainActivity so that they are not recreated each time the activity is destroyed creating persistence. 

The new file `CustomViewModelFactory` is here to unify the different factories of the view models and to allow the refactoring in `MainActivity`

## Code Coverage
image

Closes #239

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/MainActivity.kt
+++ b/app/src/main/java/com/github/se/cyrcle/MainActivity.kt
@@ -3,12 +3,15 @@ package com.github.se.cyrcle
 import CyrcleNavHost
 import android.content.pm.ActivityInfo
 import android.os.Bundle
+import android.util.Log
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
+import androidx.activity.viewModels
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.material3.Surface
 import androidx.compose.ui.Modifier
 import androidx.navigation.compose.rememberNavController
+import com.github.se.cyrcle.model.CustomViewModelFactory
 import com.github.se.cyrcle.model.address.AddressRepository
 import com.github.se.cyrcle.model.address.AddressViewModel
 import com.github.se.cyrcle.model.image.ImageRepository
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/se/cyrcle/MainActivity.kt
```diff
@@ -43,18 +45,26 @@ class MainActivity : ComponentActivity() {
 
   @Inject lateinit var authenticator: Authenticator
 
+  private val reviewViewModel: ReviewViewModel by viewModels {
+    CustomViewModelFactory { ReviewViewModel(reviewRepository) }
+  }
+  private val userViewModel: UserViewModel by viewModels {
+    CustomViewModelFactory { UserViewModel(userRepository, parkingRepository) }
```
sean-prz, line 52: you removed a new argument I just added to the userviewmodel which is the imageRepository,

It still compiles because I added a default null arg. but I can't log in anymore 

image
Easy to debug thanks to the log

Pijukai: Okay that maybe also why the CI don't work thanks i will look into it right away.

Pijukai: I think I need to rebase i branched before your PR.

sean-prz (approved):
Nice now it works.
There are other bugs that still need to be addressed about the dark mode.
Like the fact that all text inputs are cleared when switching theme, but this was the biggest bug to treat, good job.

Pijukai:
> Nice now it works. There are other bugs that still need to be addressed about the dark mode. Like the fact that all text inputs are cleared when switching theme, but this was the biggest bug to treat, good job.

The fact is this review is meant to prevent view models from being destroyed. For the Screen it is a little bit more tricky due to the nature of each screen and how they are initialized.

matttsss (approved):
All good for me, it works as intended, I could not find any new bugs and all tests are green!

matttsss:
@sean-prz I'm not sure we should start putting every text input field into a view model... This problem seem extreamly minor to me 😅

Pijukai:
Credit for you @matttsss for helping me debug a bug that was only present locally but was still annoying :wink:

sean-prz:
> @sean-prz I'm not sure we should start putting every text input field into a view model... This problem seem extremely minor to me 😅

yeah yeah that's absolutely not what I was suggesting, I was thinking about trying to use rememberSaveable instead of remember this save the state across configurations change like switching to darktheme.

matttsss:
Ohhh right I didn't know that one 👌

PR #259, opened 2024-11-20
Add Rack Type Help Screen

Armetiis:
This PR introduces a new screen that provides detailed information about different types of bike racks. This feature enhances user understanding of various rack types when adding a new parking spot.

image
image

### Changes
- Created a new composable RackTypeHelpScreen that displays a list of different rack types.
- Implemented alternating layout for rack type items (image on left/right) for visual variety.
- Integrated the new screen with the existing navigation system (clickable icon in Attribute Picker Screen).
### To do in the future:
 Add actual images for each rack type (currently using a placeholder). Ideally, I thought it could be nice to try taking the pictures ourserlves on campus (that way we're sure we don't stumble upon copyright issues)

### Coverage
image


Pijukai:
I agree with your idea of taking pictures ourselves. That is the best way to do good pictures, test the app functionality and  of course like you mentionned avoid copyright issue.

app/src/main/java/com/github/se/cyrcle/ui/addParking/attributes/RackInfo.kt
```diff
@@ -0,0 +1,126 @@
...
+@Composable
+fun RackText(rackType: ParkingRackType, modifier: Modifier = Modifier) {
+  Text(
+      text = stringResource(getRackDescriptionResource(rackType)),
+      style = MaterialTheme.typography.bodySmall.copy(fontSize = 12.sp),
+      textAlign = TextAlign.Justify,
+      modifier = modifier.testTag("rackDescription_${rackType.name}"))
+}
+
+fun getRackDescriptionResource(rackType: ParkingRackType): Int {
```
Pijukai, line 115: There is a documentation missing here.

app/src/main/java/com/github/se/cyrcle/ui/addParking/attributes/AttributesPicker.kt
```diff
@@ -96,19 +102,19 @@ fun AttributesPicker(
...
 
   val location = mapViewModel.selectedLocation.collectAsState().value!!
   LaunchedEffect(location) { addressViewModel.search(location.center) }
   val suggestedAddress = addressViewModel.address.collectAsState().value
-  LaunchedEffect(suggestedAddress) { title.value = suggestedAddress.displayRelevantFields() }
+  LaunchedEffect(suggestedAddress) {
+    if (title.value.isEmpty()) title.value = suggestedAddress.displayRelevantFields()
+  }
 
   fun onSubmit() {
```
Pijukai, line 119: There is a missing documentation here as well

Armetiis: what do you mean missing documentation? If you're talking about the onSubmit,it's just defined here to be used as a lambda. It's basically a variable. In any case, I did not make this code and am therefore not responsible for its documentation

Pijukai: Ohhh i didn't see that you weren't the one who wrote this code my bad.

Pijukai (approved):
Well done @Armetiis this is a very good PR and the info page about the rack types is very clear. Thanks for the fixes. The PR is good to go as is however i have an idea that you could quickly add. How about adding this page in the Card also for each parking to inform the user when they take a look at a parking?  Anyway Approved :+1:

app/src/main/java/com/github/se/cyrcle/ui/addParking/attributes/AttributesPicker.kt
```diff
@@ -180,11 +186,24 @@ fun AttributesPicker(
...
-                    )
+                    Row(
+                        verticalAlignment = Alignment.CenterVertically,
+                        modifier = Modifier.fillMaxWidth()) {
+                          EnumDropDown(
+                              options = ParkingRackType.entries.toList(),
+                              selectedValue = rackType,
+                              label = stringResource(R.string.attributes_picker_rack_type_label),
+                              modifier = Modifier.weight(1f).fillMaxWidth())
+                          Icon(
```
andrepeiry, line 197: I think that such an Icon can and will be reused later in our app, would you be so kind as to turn it into an Atom ?

andrepeiry (commented):
Everything seems to be of good quality, modular and reusable. Just a little thought on the outside of your class for the Info Icon that can be turned into an Atom and that would be perfect for me !

app/src/main/java/com/github/se/cyrcle/ui/addParking/attributes/AttributesPicker.kt
Armetiis: I think we can do that when we plan to reuse it (for example as Julien mentionned, in the Parking Details Screen). Does that sound right to you?

Armetiis:
> How about adding this page in the Card also for each parking to inform the user when they take a look at a parking?
@Pijukai we can do that in a future PR dedicated to that, and other minor changes to the detail screen (I just noticed the parking description is not shown in the card screen)

PR #261, opened 2024-11-21
Fix and Enhance Mock Tests for Firestore Repositories

Armetiis:
This Pull Request addresses and resolves issues with the mock tests for our Firestore repositories, significantly improving their accuracy and coverage.

### Changes
Complete rework of the tests for the following Firestore repositories:
- ParkingRepositoryFirestore
- UserRepositoryFirestore
- ReviewRepositoryFirestore

### Improvements
1. **Test Accuracy**: The primary goal was to correct the syntax errors in our tests, ensuring they accurately assess our repository implementations.
2. **Coverage Increase**: As a beneficial side effect, these changes have substantially increased our test coverage:

   | Repository | Previous Coverage | New Coverage |
   |------------|-------------------|--------------|
   | ParkingRepositoryFirestore | 63% | 95% |
   | UserRepositoryFirestore | 53% | 92% |
   | ReviewRepositoryFirestore | 51% | 95% |

	**Overall Impact**: Our total coverage has improved from 81% to 84%.

### Implementation Details
The tests now utilize an adapted version of the syntax provided by Florian in response to the private Ed question I posted on the class forum two weeks ago.

### Visual Proof
Coverage screenshots have been attached to visually demonstrate the improvements:
- Overall coverage
image
- ParkingRepositoryFirestore coverage
image 
- UserRepositoryFirestore coverage
image
- ReviewRepositoryFirestore coverage
image

These changes not only fix the existing issues but also provide a more robust testing framework for our Firestore repositories.


app/src/test/java/com/github/se/cyrcle/model/user/UserRepositoryFirestoreTest.kt
```diff
@@ -55,92 +70,212 @@ class UserRepositoryFirestoreTest {
...
   @Test
-  fun getAllUsers_returnsCorrectValues() {
-    // Create a TaskCompletionSource to manually complete the task
-    val taskCompletionSource = TaskCompletionSource<QuerySnapshot>()
+  fun getUidTest_throwsException_whenUidIsEmpty() {
+    `when`(mockFirebaseAuth.currentUser).thenReturn(null)
+    try {
+      userRepositoryFirestore.getUid()
+    } catch (e: Exception) {
+      assert(e.message == "User not signed in")
```
sean-prz, line 78: If no exception is thrown, we don't enter the catch block. So we don't assert that the exception has been thrown

app/src/test/java/com/github/se/cyrcle/model/user/UserRepositoryFirestoreTest.kt
```diff
@@ -55,92 +70,212 @@ class UserRepositoryFirestoreTest {
...
+    `when`(mockDocumentReference.get()).thenReturn(taskCompletionSource.task)
 
     userRepositoryFirestore.getUserById(
         userId = user.public.userId,
-        onSuccess = { assert(it == user) },
-        onFailure = { fail("Expected success but got failure") })
-    verify(timeout(100)) { mockDocumentReference.get() }
-    verify(timeout(100)) { mockDocumentSnapshot.toObject(User::class.java) }
+        onSuccess = { fail("Expected failure but got success") },
+        onFailure = { assert(true) })
```
sean-prz, line 116: Same here, the test is useless, if onFailure is not launched the test still passes

app/src/test/java/com/github/se/cyrcle/model/user/UserRepositoryFirestoreTest.kt
```diff
@@ -55,92 +70,212 @@ class UserRepositoryFirestoreTest {
...
+    // Arrange
+    val exception = Exception("Test exception")
+    val taskCompletionSource = TaskCompletionSource<DocumentSnapshot>()
+    taskCompletionSource.setException(exception)
+    `when`(mockDocumentReference.get()).thenReturn(taskCompletionSource.task)
+
+    // Act
+    userRepositoryFirestore.getUserById(
+        userId = user.public.userId,
+        onSuccess = { fail("Expected failure but got success") },
```
sean-prz, line 137: this test too doesn't assert that OnFailure was called, right ?

sean-prz (changes requested):
I might be wrong but it seems a lot of tests don't actually test anything.
I have only looked at a small portion of the tests but it seems this is a reccurent problem, so let's fix this and I'll review again 
do you agree?

app/src/test/java/com/github/se/cyrcle/model/user/UserRepositoryFirestoreTest.kt
Armetiis: You're right

app/src/test/java/com/github/se/cyrcle/model/user/UserRepositoryFirestoreTest.kt
Armetiis: If on failure is not launched then on sucess is? Is there a case where none is launched? Maybe I am confused...

sean-prz: We can't assume that, this is the goal of these unit tests. To make sure we implemented it well and that there is no path that doesn't take any of the callbacks.

Also it's pretty hard to setup the mock correctly so that they call the callbacks.

Armetiis:
@sean-prz I'll push a fix for this. You make a very good point, I am sorry I overlooked this

sean-prz (approved):
Thanks for the changes ! And thank you for lifting our coverage up !

LouisJamesVasseur (approved):
Fantastic job on fixing this! I see nothing to add except to thank you for increasing our coverage :)

PR #267, opened 2024-11-21
fix: transformPathToUrl crashes the app with old user

sean-prz:
### What

This PR fixes the following fatal bug :
`TransformPathToUrl` was called with a null `Path` for user that set up their account with the old M2 versions of the app. 

The function wasn't handling that case and it just made the app crash, now the function will just skip and return instantly if called for an old deprecate user


Armetiis (approved):
👍

matttsss (approved):
:+1:

PR #269, opened 2024-11-22
File IO

### LINKED ISSUE #262: Interface for file IO

### LINKED ISSUE #271: Groundwork for offline mode

andrepeiry:
# Summary

This PR is the ground work for the offline part of the application.

# Content

- A preference storage, using a Data Store for bribes of information that need to be stored and access quickly, such as the theme of the app (for example)
- A generic Proto Serializer, if we want to store basic classes (i.e. data class with primitive variables)
- A map tile storage, for user to save in advance the map of a certain location

# Potential improvements for later

- Add Enums to the Proto Serializer
- Better documentation for the rest of the team, notably for Proto

# Close

Close #262 
Close #271 

# Code Coverage

Code Coverage as of commit 735a027


app/src/main/java/com/github/se/cyrcle/io/MapTileStorage.kt
```diff
@@ -0,0 +1,46 @@
...
+ * @param zoomLevel The zoom level of the tile.
+ * @param x The x-coordinate of the tile.
+ * @param y The y-coordinate of the tile.
+ * @param tileData The byte array of the tile data.
+ * @throws IOException If an I/O error occurs.
+ */
+@Throws(IOException::class)
+fun storeTile(zoomLevel: Int, x: Int, y: Int, tileData: ByteArray) {
+  val tileDir = File("$CACHE_DIR/$zoomLevel/$x")
+  if (!tileDir.exists()) {
```
AlexandreBeeDeeBop, line 21: You can simplify with :
File("$CACHE_DIR/$zoomLevel/$x").apply { mkdirs() }
which does nothing if the directory already exists.

app/src/main/java/com/github/se/cyrcle/io/MapTileStorage.kt
```diff
@@ -0,0 +1,46 @@
+package com.github.se.cyrcle.io
+
+import java.io.File
+import java.io.IOException
+
+private const val CACHE_DIR = "./cache/map"
```
AlexandreBeeDeeBop, line 6: did you consider using android's cache directory ?

app/src/test/java/com/github/se/cyrcle/io/Settings.kt
```diff
@@ -0,0 +1,61 @@
...
+    result = 31 * result + heightInCm.hashCode()
+    result = 31 * result + weightInMg.hashCode()
+    result = 31 * result + isDeveloper.hashCode()
+    return result
+  }
+
+  class Serializer :
+      ProtoGenericSerializer<Settings>(Settings("John Doe", 42, 420000, 4.2f, 4.425397, true)) {
+    override suspend fun readFrom(input: InputStream): Settings {
+      return Settings(
```
AlexandreBeeDeeBop, line 43: I think you could add some error handling here just in case. Either log the error or fallback to defaultValue or something like that maybe ?

app/src/test/java/com/github/se/cyrcle/io/Settings.kt
```diff
@@ -0,0 +1,61 @@
+package com.github.se.cyrcle.io
+
+import java.io.InputStream
+import java.io.OutputStream
+
+data class Settings(
```
AlexandreBeeDeeBop, line 6: Will the values that you'll actually put in later need a validation function or do we expect that to be done previously elsewhere ?

app/src/main/java/com/github/se/cyrcle/io/MapTileStorage.kt
andrepeiry: Yes, but no, as the goal of this is to not be a cache for the application, but rather some content that should be store directly (think of downloading a part of the worlds). But what your comment makes me realize is that the directory name isn't well define, as this is no cache. This will be address

app/src/test/java/com/github/se/cyrcle/io/Settings.kt
andrepeiry: What do you mean by *validation* ? Like checking is a number shouldn't be negative ? This would done at class instantiation (construction)
If you're speaking only for this class, you'll notice that this one is simply for tests, so there is no point in having data validation for it (other that for tests)

app/src/main/java/com/github/se/cyrcle/io/PreferenceStorage.kt
```diff
@@ -0,0 +1,41 @@
...
+
+import android.content.Context
+import androidx.datastore.core.DataStore
+import androidx.datastore.preferences.core.Preferences
+import androidx.datastore.preferences.core.edit
+import androidx.datastore.preferences.preferencesDataStore
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.map
+
+val Context.preferencesDataStore: DataStore<Preferences> by preferencesDataStore(name = "settings")
```
LouisJamesVasseur, line 11: I think the definition of a "preference in the comments of this class could be good. I get what it is, but I feel like it could be made clearer in your code and in your PR for the rest of the group who is not reviewing you :)

app/src/main/java/com/github/se/cyrcle/io/ProtoGenericSerializer.kt
```diff
@@ -0,0 +1,171 @@
...
+/**
+ * A class to easily serialize and deserialize class with basic variables inside. An example is
+ * available in the test folder, under Settings.
+ */
+abstract class ProtoGenericSerializer<T>(
+    override val defaultValue: T,
+) : Serializer<T> {
+
+  companion object {
+    private fun writeNByte(output: OutputStream, byteArray: ByteArray) {
```
LouisJamesVasseur, line 19: This is missing the plural (quite important since we write multiple byte*s* :) )

app/src/test/java/com/github/se/cyrcle/io/Settings.kt
LouisJamesVasseur: I agree that error handling could be useful since serialization has been a common problem case in our project

app/src/main/java/com/github/se/cyrcle/io/MapTileStorage.kt
```diff
@@ -0,0 +1,46 @@
...
+ * @throws IOException If an I/O error occurs.
+ */
+@Throws(IOException::class)
+fun storeTile(zoomLevel: Int, x: Int, y: Int, tileData: ByteArray) {
+  val tileDir = File("$CACHE_DIR/$zoomLevel/$x")
+  if (!tileDir.exists()) {
+    tileDir.mkdirs()
+  }
+  val tileFile = File(tileDir, "$y.$EXTENSION_NAME")
+  if (tileFile.exists()) return // Tile should change, there is no point in updating them
```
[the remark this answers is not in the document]
andrepeiry: Indeed, this seems to be more simpler and less error-prone. I'll even modify it to allow update with some overwrite argument, so that usage is more flexible and logic more readable.

app/src/main/java/com/github/se/cyrcle/io/MapTileStorage.kt
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Load a tile from the specified zoom level and coordinates.
+ *
+ * @param zoomLevel The zoom level of the tile.
+ * @param x The x-coordinate of the tile.
+ * @param y The y-coordinate of the tile.
+ * @return The byte array of the tile data, or null if the tile does not exist.
+ * @throws IOException If an I/O error occurs.
+ */
+@Throws(IOException::class)
```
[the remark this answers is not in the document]
andrepeiry: I think it is an artifact of a previous iteration of code, it is indeed not useful anymore, I'll remove it

app/src/main/java/com/github/se/cyrcle/io/PreferenceStorage.kt
andrepeiry: Indeed, some documentation would be require for this, I'll be on it !

app/src/main/java/com/github/se/cyrcle/io/PreferenceStorage.kt
```diff
@@ -0,0 +1,41 @@
...
+class PreferenceStorage(private val context: Context) {
+
+  /**
+   * Write a preference associated with [preferenceKey] to the data store. If the value is already
+   * present, it will be overwritten.
+   *
+   * @param preferenceKey The key of the preference to write.
+   * @param value The value to associate with [preferenceKey].
+   */
+  suspend fun <T> writePreference(preferenceKey: Preferences.Key<T>, value: T) {
```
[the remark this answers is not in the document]
andrepeiry: Sadly, there is no easy way to ensure that no one overwrite anything. I thought of having a list right above it, but I already know it won't be up-to-date with everybody.
I also thought of having an Enum to solve this, but at this point, it was more obfuscation than it would be helping anybody.
I guess it will be up to the team not to mess this.

app/src/test/java/com/github/se/cyrcle/io/Settings.kt
```diff
@@ -0,0 +1,61 @@
...
+    if (heightInCm != other.heightInCm) return false
+    if (weightInMg != other.weightInMg) return false
+    if (isDeveloper != other.isDeveloper) return false
+
+    return true
+  }
+
+  override fun hashCode(): Int {
+    var result = userName.hashCode()
+    result = 31 * result + age
```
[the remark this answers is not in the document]
andrepeiry: I had to rewrite the `.equals` function, and as it should, the warning *Equals redefined without redefining `hashCode`* showed up. Copilot cooked me something in 3.4 seconds and that was it. As it is a test class and that it isn't even used, this comment won't be addressed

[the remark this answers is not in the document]
LouisJamesVasseur: Fair enough in that case, this is nonblocking for me

app/src/main/java/com/github/se/cyrcle/io/PreferenceStorage.kt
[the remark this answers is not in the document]
LouisJamesVasseur: I see, and understand that these things are sometimes unavoidable. We'll have to be careful when implementing offline mode

app/src/main/java/com/github/se/cyrcle/io/PreferenceStorage.kt
LouisJamesVasseur: Thanks!

LouisJamesVasseur (approved):
Great work addressing all my comments, and great job on the PR! Approved!

app/src/main/java/com/github/se/cyrcle/io/MapTileStorage.kt
AlexandreBeeDeeBop: Ok I understand.

LouisJamesVasseur (approved):

AlexandreBeeDeeBop (approved):

PR #280, opened 2024-11-22
Check that users exist for sign in and account creation

### LINKED ISSUE #238: Make account creation backend

matttsss:
## What is this PR? 
This PR implements checks for the user's existence in our database

This is reflected in the UI in the following way:
* If the user attempts to create an account:
    * If his account exists already, toast and redirect him to sign in
    * If it doesn't exist, create the account, log in and go to map
* If the user attempts to sign in:
    * If his account exists, log him in and go to map
    * If it doesn't exist, toast and stay in sign in screen

**Footnote:** It also removes the `null` default argument from the `UserViewModel`

## How?
A new function was added to the `UserRepository`, `userExists` that takes a callback `(Boolean) -> Unit` and a Failure callback.

### Code coverage:
- Overall coverage at 82%
- UserRepositoryFirestore at 93%
- UserViewModel at 80%
- SignInScreen at 93%

Closes #238

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/model/user/UserRepositoryFirestore.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/user/UserRepositoryFirestore.kt
@@ -68,6 +68,22 @@ constructor(private val db: FirebaseFirestore, private val auth: FirebaseAuth) :
         .addOnFailureListener(onFailure)
   }
 
+  override fun userExists(
+      user: User,
+      onSuccess: (Boolean) -> Unit,
+      onFailure: (Exception) -> Unit
+  ) {
+    db.collection(collectionPath)
+        .whereEqualTo("userId", user.public.userId)
+        .whereEqualTo("username", user.public.username)
+        .get()
+        .addOnSuccessListener {
+          Log.d("UserRepositoryFirestore", "User exists: ${it.size()}")
+          onSuccess(!it.isEmpty)
+        }
+        .addOnFailureListener(onFailure)
+  }
+
   override fun addUser(user: User, onSuccess: () -> Unit, onFailure: (Exception) -> Unit) {
     if (user.details == null) {
       onFailure(Exception("User details are required"))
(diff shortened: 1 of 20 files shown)


sean-prz (commented):
Weird, it doesn't let me log-in but I have an account in the firestore ? is this expected.
The app want me to sign-up

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
```diff
@@ -89,6 +89,21 @@ class UserViewModel(
...
+   *
+   * @param user the user to check
+   * @param onSuccess the callback to call with a boolean indicating if the user exists
+   */
+  fun doesUserExist(user: User, onSuccess: (Boolean) -> Unit) {
+    userRepository.userExists(
+        user,
+        onSuccess = onSuccess,
+        onFailure = {
+          Log.e("UserViewModel", "Failed to check if user exists: ${user.public.userId}", it)
```
Pijukai, line 103: I don't know if it is a log of debug and if it is intended to stay. But if it is, wouldn't it be more useful if they were both in the `userExist()` function in the repository so that we are sure that there is a problem with the repo if need be ?

app/src/main/java/com/github/se/cyrcle/model/user/UserRepositoryFirestore.kt
```diff
@@ -68,6 +68,22 @@ constructor(private val db: FirebaseFirestore, private val auth: FirebaseAuth) :
...
   }
 
+  override fun userExists(
+      user: User,
+      onSuccess: (Boolean) -> Unit,
+      onFailure: (Exception) -> Unit
+  ) {
+    db.collection(collectionPath)
+        .whereEqualTo("userId", user.public.userId)
+        .whereEqualTo("username", user.public.username)
```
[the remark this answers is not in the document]
matttsss: Yes that's correct thanks a lot! The `SignInScreen` passes the temporary `User` that only have the default info given by Firestore. The `username` check might be removed since it involves useless loads from and to Firestore.

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
matttsss: Right, I moved the `Log` down to the `UserRepository` for consistency but left the `onFailure` callback in case it will be needed later.

Pijukai:
image
It seems that the atom has quite an annoying formatting with very long username , we need to make an issue to fix this next sprint. I don't think you need to adress it matteo since it is not the goal of your PR.

sean-prz (approved):
Nice, everything works now, sorry you had through the painful changes I lazily avoided :/

PR #282, opened 2024-11-23
Prevent users from adding thin and long parkings

### LINKED ISSUE #281: Add a limit for width and height of parkings
Currently users have an area limit for their parkings. However, this doesn't prevent them from adding super long parkings (e.g width = 1m and height = 1000m)

The solution is to add a second check for the width and the length of the parking

Armetiis:
This PR addresses issue #281 by implementing a limit on both the width and height of parkings, in addition to the existing area limit. This change prevents users from adding extremely long or thin parkings while maintaining the current area restrictions.

### Changes
- Added a new constant `PARKING_MAX_SIDE_LENGTH` in `ParkingViewModel.kt`.
- Modified `MapViewModel.kt` to include a new `isLocationValid` flow that checks both area and side length constraints
- Updated `Parking.kt` to include a new `computeHeightAndWidth()` function. Also added null checks to `toPolygon()` function.
- Updated UI components to use the new `isLocationValid` state instead of the previous isAreaTooLarge.

### Testing
Manually tested the new validation logic by attempting to create parkings with various dimensions

### Notes
- The maximum side length is set to 50 meters, but this value can be adjusted if needed.

### Coverage
image

-- -
Closes #281

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
@@ -6,6 +6,7 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.ViewModelProvider
 import com.github.se.cyrcle.model.parking.Location
 import com.github.se.cyrcle.model.parking.PARKING_MAX_AREA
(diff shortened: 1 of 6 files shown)


matttsss (approved):
Simple and effective, not much to say!

PR #284, opened 2024-11-23
Display all parkings in List Screen with pin distinction

### LINKED ISSUE #283: Display pinned parkings too in "all parkings" in List Screen
### Current Behavior:
In the "All parkings" section of the List Screen, only non-pinned parkings are displayed. This causes pinned parkings to disappear from the user's view.
### Desired Behavior:
Display all parkings in the "All parkings" section, including both pinned and non-pinned, while maintaining a clear visual distinction between the two types.

Armetiis:
This PR addresses the issue of pinned parkings disappearing from the "All parkings" section in the List Screen. It improves user experience by showing all parkings while maintaining a clear visual distinction for pinned items.

### Changes

1. Modified `ListScreen.kt` to display all parkings in the main list, regardless of pin status.
2. Added a unique key for pinned parkings to prevent app crashes due to duplicate keys in LazyColumn.
```kotlin
key = { parking -> parking.uid + "pin" }
```
3. Removed the filter that excluded pinned parkings from the main list

### Notes

- This change maintains the separate "Pinned parkings" section at the top of the list.
- Users can now see all parkings in one view, with pinned items appearing twice (in both sections).
### Coverage
The rules are the rules, but it feels weird to upload the coverage for a 2 line fix...
image

Closes #283

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
@@ -102,6 +102,7 @@ fun SpotListScreen(
           FilterHeader(parkingViewModel = parkingViewModel)
           val listState = rememberLazyListState()
           LazyColumn(state = listState, modifier = Modifier.testTag("SpotListColumn")) {
+            // Pinned parking spots if any (includes titles and dividers)
             if (pinnedParkings.isNotEmpty()) {
               item {
                 Text(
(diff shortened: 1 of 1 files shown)


matttsss (commented):
It's a bit picky but could it be possible to move the pin marker under the distance?
image
Since when we pin, the whole content of the card is shifed down to make room for the pin icon when placing it under wouldn't shift the card content?

Armetiis:
> Since when we pin, the whole content of the card is shifed down to make room for the pin icon when placing it under wouldn't shift the card content?

I hear you. I pushed a fix, tell me what you think. Funny to note that between this small fix and ktfmt, a 4 line PR became a 74 lines one 😂

Anyway I think it's a good remark, I also like it better like this

matttsss (approved):
Nice! Thanks for the changes! I like it!

sean-prz (approved):
You made pinning parking a realy smooth experience. 
I keep pinning parking... so satisfying ...  Thanks

PR #285, opened 2024-11-23
Upload&Download Pictures for parkings

### LINKED ISSUE #291: Large pictures makes the app crash on upload
The function in charge of compressing the image before sending it to the database struggle with image of large size. 
This function should be reworked to handle that case

sean-prz:
_Closes #286 #287_
### Implementing the Image Repository into the parking viewmodel.

The parking viewmodels now offers to the UI two new functions :
- `uploadImage` : upload our pictures to our cloud storage and update the parking object to contains the path to that picture.
- `loadSelectedParkingImages` : get all paths stored into the selected parking object and transform them into a list of URL that is passed via a state `selectedParkingImagesUrls` to the UI 

### Modifying the UI
On the details screen the following changes have been made : 
- Reuse Alexandre's code for the image Picker used for profile pictures. 
-  Adds a confirmation screen before uploading the picture.
- The details screen now observe a state of URLS to recompose the list of image each time a new URLS has been transformed.

#### Being able to see pictures of the spot is an important feature that helps the user to locate the spots and easily see if it fits their expectation

-- -
#### Global Code coverage 
image

#### new Code coverage 
- ParkingViewModel at 92%
- UI ParkingDetailsScreen at 83%

-- -

### What's next 
I'll take any suggestions on what needs to be done to polish this feature.
What can be done to improve this new feature : 
- Making a picture clickable to expand it.
- allow user to delete their pictures
- Preventing users from adding more than 5 pictures per spot ?
- allowing User to report pictures
- fix #291


matttsss (commented):
While reviewing I found a small bug... When I tried adding a photo in the ParkingDetailsScreen, the images wouldn't show up... I had to relaunch the app to be able to see the images... Maybe some kind of caching issue?
From the logs, the parking managed to update successfully but the images were not displayed.

sean-prz:
> While reviewing I found a small bug... When I tried adding a photo in the ParkingDetailsScreen, the images wouldn't show up... I had to relaunch the app to be able to see the images... Maybe some kind of caching issue? From the logs, the parking managed to update successfully but the images were not displayed.

Oh.. right, it's because I only fetch the URL's once when arriving on the screen.
We can fix this by observing also the selectedParking, so that when it changes (new pic added) we refetch the URL's 
I've put the selectedParking in the trigger of the launchedEffect an it seems to work now.

sean-prz:
Made some tests to stay at 82%

matttsss:
Right I can now see the images that I upload! However it seems like images too big crash the program 😅

sean-prz:
> Right I can now see the images that I upload! However it seems like images too big crash the program

Well that has to be because of my compressing function, it downscale the images 5% by 5% until the image size is below 300kb, but I guess if at 5% of the original Resolution we're still above 300kb, the edge case is not well handled, let me try to push a fix

matttsss (approved):
Nice changes, the image load is still a bit slow, could defenitively be worked on for a later PR but it's nice to be able to add pictures!

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
```diff
@@ -193,41 +234,38 @@ fun ParkingDetailsScreen(
...
-                                      Modifier.size(170.dp)
-                                          .padding(4.dp)
-                                          .testTag("ParkingImage$index"),
-                                  contentScale = ContentScale.Crop)
+                            items(imagesUrls) { url ->
+                              Image(
+                                  painter = rememberAsyncImagePainter(url),
+                                  contentDescription =
+                                      stringResource(
+                                          R.string.view_profile_screen_profile_picture), // FIXME
```
Armetiis, line 256: Here you left a fixme. Note that you don't need to use a string resource for the content descriptions

app/src/main/res/values/strings.xml
```diff
@@ -88,6 +88,9 @@
     <string name="card_screen_add_review">Add A Review</string>
     <string name="card_screen_report">Report</string>
     <string name="card_screen_no_image">No images for this parking yet</string>
+    <string name="card_screen_confirm_upload">Do you want to add this image to the parking spot ? </string>
+    <string name="card_screen__accept_upload">Yes</string>
+    <string name="card_screen__cancel_upload">No</string>
```
Armetiis, line 93: We already have general resources for Yes and No, no need to define them again for this specific purpose

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
```diff
@@ -69,6 +78,38 @@ fun ParkingDetailsScreen(
...
   val toastMessage = stringResource(R.string.sign_in_to_add_favorites)
+  // === States for the images ===
+  var newParkingImageLocalPath by remember { mutableStateOf("") }
+  val showDialog = remember { mutableStateOf(false) }
+  val imagesUrls by parkingViewModel.selectedParkingImagesUrls.collectAsState()
+  // === === === === === === ===
+
+  LaunchedEffect(Unit, selectedParking) {
+    // On first load of the screen, request the images
+    // This will update the imagesUrls state and trigger a recomposition
```
Armetiis, line 89: Why do you have a launched effect of unit and selected parking? Isn't just selected parking enough?

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsAlertDialog.kt
```diff
@@ -0,0 +1,71 @@
...
+    onDismiss: () -> Unit,
+    onAccept: () -> Unit,
+    newParkingImageLocalPath: String,
+) {
+  AlertDialog(
+      modifier = Modifier.testTag("parkingDetailsAlertDialog"),
+      onDismissRequest = { onDismiss() },
+      title = { Text(stringResource(R.string.card_screen_confirm_upload)) },
+      text = {
+        Column(
```
Armetiis, line 37: I don't think you need to wrap the image in a column. You can just add a container color parameter to the Alert Dialog to use surface bright

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
```diff
@@ -436,6 +439,67 @@ class ParkingViewModel(
...
+    val destinationPath = "parking/${selectedParking.uid}/${selectedParking.images.size}"
+    imageRepository.uploadImage(
+        context = context,
+        fileUri = imageUri,
+        destinationPath = destinationPath,
+        onSuccess = {
+          selectParking(selectedParking.copy(images = selectedParking.images.plus(destinationPath)))
+          parkingRepository.updateParking(
+              selectedParking,
+              { onSuccess },
```
Armetiis, line 494: I don't see the purpose of having an onSuccess parameter for now. This is not a problem, we could need it in the fututre. However here it should be `onSucess()`, because `onSuccess` is not a correct syntax for invoking the function

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
```diff
@@ -193,41 +234,38 @@ fun ParkingDetailsScreen(
...
-                                          .padding(4.dp)
-                                          .testTag("ParkingImage$index"),
-                                  contentScale = ContentScale.Crop)
+                            items(imagesUrls) { url ->
+                              Image(
+                                  painter = rememberAsyncImagePainter(url),
+                                  contentDescription =
+                                      stringResource(
+                                          R.string.view_profile_screen_profile_picture), // FIXME
+                                  modifier = Modifier.size(100.dp),
```
Armetiis, line 257: I think here maybe instead of setting the size of the image, we could rather have a fixed height for the lazyrow and have the images fill max height. Honesty not a concern for this PR, as I think the UI could use a bit more work (e.g. when there are images, the add button looks a bit odd hanging there alone). Just taking the chance to mention it and confirm that you agree with me (or not)

Armetiis (commented):
Very good job, very nice feature and clean implementation. Your code includes a lots of comments which made it very easy to understand, so thank you for that. I left you some remarks in the code. The one that bothers me the most would be the onSuccess call in the parkingViewModel

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
sean-prz: fixed

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
sean-prz: I'll add a comment to make it clearer !

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
sean-prz: I don't see it either, but it's nice to have a callback for everything that is async in case we ever need it, I'll fix the onSuccess()

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
sean-prz: Sure I'll add this to the list of things to be done

Armetiis (approved):
Thanks for the changes!

Issue #274, opened 2024-11-22
Refactor Authenticator to provide callbacks


matttsss:
In the current state, the `Authenticator` only provides buttons which constrains the UI a lot.
The ideal `Authenticator` class should provide button callbacks instead of buttons.

PR #288, opened 2024-11-23
Update Google SignIn API

### LINKED ISSUE #277: Remove deprecated `GoogleSignIn` api
In the current state, the application uses a deprecated API to Sign In via Google.
The ideal app should use the newer Google API.

matttsss:
## What is this PR?

This PR aims to remove the deprecated version of the Google SignIn API that we used. It now uses the new API that does not launch any Intents.

## Future work (#274):
With this new API, the `Authenticator` system is not necesary since we don't need compose functions to authenticate.
My next PR will be aimmed at integrating authentication code within the `UserViewModel`.

### Test Coverage:
image

Closes #277

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -166,6 +166,11 @@ dependencies {
     implementation(libs.compose.preview)
     debugImplementation(libs.compose.tooling)
 
+    implementation(libs.credentials)
+    implementation(libs.credentials.play.services.auth)
+    implementation(libs.android.googleid)
+
+
 
     // ---------------------------------------------------
     // -------------------  PLUGINS  ---------------------
@@ -173,7 +178,6 @@ dependencies {
 
     // ---------------- Firebase  ------------------
     implementation(libs.google.services)
-    implementation(libs.play.services.auth)
 
     val firebaseBom = platform(libs.firebase.bom)
     implementation(firebaseBom)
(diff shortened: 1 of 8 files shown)


sean-prz (approved):
Everything still works, I tested with my account/new account and as a guest. 

For the code I can't provide any feedback, it's just too in depth on the topic of authentification. It's just useless for me to dive as deep as you on this.

It is a bit hard to follow the code.
When I read it, I have to jump from AuthenticatorImpl.kt  to Authenticator.kt  to check the default button, then back to AuthenticatorImpl to see the two params passed  then further down the code to see the actual params, which themselves are defined in another function.
Would be nice if you could write a summary or draw a graph so we get a grasp of what's happening. 

I will approve regardless because I don't think any one but the reviewer will have to understand this code

matttsss:
@sean-prz Yes the process is a bit complicated... I am currently working on simplifying it by integrating the `Authenticator` to the `UserViewModel` to get one API

Armetiis (approved):
To be completely honest I have nothing to say. Works good. I'd like to comment on the code but I must admit that it's a bit too niche and as Sean implied, the research necessary to understand this code is quite tedious to be done by all the reviewers in addition to you. 

I still appreciate that you left good documentation and comments to help potential future people maintain the code

PR #290, opened 2024-11-23
Optimize Favorite Parking State Management in UserViewModel

### LINKED ISSUE #289: Synchronization of Favorite Parkings State with User's favorite parking field
### Description:
Currently, the favorite parkings state is not in sync with the current user's selected parkings. The functions `addFavoriteParkingToSelectedUser` and `removeFavoriteParkingFromSelectedUser` in the `UserViewModel` do not update this state, which implies that the UI has to fetch every single time it adds or removes a favorite.

### Proposed Solution:
1. Make the `getSelectedUserFavoriteParking` function private and call it only when the user is selected.
2. Use the favorite parking state only in the UI, which is automatically updated in the `addFavoriteParkingToSelectedUser` and `removeFavoriteParkingFromSelectedUser` functions.

Armetiis:
This PR addresses the issue of favorite parking state synchronization with the current user's selections. It improves the `UserViewModel` to handle state updates more efficiently and reduces unnecessary data fetching.

### Changes
1. Modified `UserViewModel`:
   - Updated `addFavoriteParkingToSelectedUser` and `removeFavoriteParkingFromSelectedUser` to accept a `Parking` object instead of a parking ID.
   - Made `getSelectedUserFavoriteParking` private, now called only upon user sign-in.
   - Implemented automatic state updates within add and remove functions.

2. Updated UI components:
   - Removed redundant fetching operations.
   - Utilized the favorite parking state directly from `UserViewModel`.

3. Adapted unit tests to reflect these changes.

### Sign in process
In orange, you can see the new steps that were added to the sign in process diagram
image

### Benefits
- Improved state management efficiency.
- Reduced unnecessary data fetching operations.
- Better code maintanability

### Testing
- Updated existing tests to accommodate new function signatures and behavior.
- Manually verified correct state updates by interacting with all screens that use this feature

### Coverage
image

Closes #289

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
@@ -130,20 +130,11 @@ fun ParkingDetailsScreen(
                                     if (userSignedIn.value) {
                                       if (isFavorite) {
                                         userViewModel.removeFavoriteParkingFromSelectedUser(
-                                            selectedParking.uid)
-                                        Toast.makeText(
-                                                context,
-                                                "Removed From Favorites!",
-                                                Toast.LENGTH_SHORT)
-                                            .show()
+                                            selectedParking)
                                       } else {
                                         userViewModel.addFavoriteParkingToSelectedUser(
-                                            selectedParking.uid)
-                                        Toast.makeText(
-                                                context, "Added To Favorites!", Toast.LENGTH_SHORT)
-                                            .show()
+                                            selectedParking)
                                       }
-                                      userViewModel.getSelectedUserFavoriteParking()
                                     } else {
                                       Toast.makeText(context, toastMessage, Toast.LENGTH_SHORT)
                                           .show()
--- a/app/src/test/java/com/github/se/cyrcle/model/user/UserViewModelTest.kt
+++ b/app/src/test/java/com/github/se/cyrcle/model/user/UserViewModelTest.kt
@@ -68,14 +68,15 @@ class UserViewModelTest {
     // Check if the user was fetched from the repository
     verify(userRepository).getUserById(eq("user1"), any(), any())
   }
-  // Check that the user returned is the correct one
-  // and that onSuccess is called.
+
+  @Suppress("UNCHECKED_CAST")
   @Test
   fun getUserBydWithCallbackTest() {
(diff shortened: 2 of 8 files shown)


app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
```diff
@@ -130,20 +130,11 @@ fun ParkingDetailsScreen(
                                     if (userSignedIn.value) {
                                       if (isFavorite) {
                                         userViewModel.removeFavoriteParkingFromSelectedUser(
-                                            selectedParking.uid)
-                                        Toast.makeText(
-                                                context,
-                                                "Removed From Favorites!",
-                                                Toast.LENGTH_SHORT)
-                                            .show()
+                                            selectedParking)
```
LouisJamesVasseur, line 133: Why were the toasts removed?

app/src/test/java/com/github/se/cyrcle/model/user/UserViewModelTest.kt
```diff
@@ -68,14 +68,15 @@ class UserViewModelTest {
     // Check if the user was fetched from the repository
     verify(userRepository).getUserById(eq("user1"), any(), any())
   }
-  // Check that the user returned is the correct one
-  // and that onSuccess is called.
+
+  @Suppress("UNCHECKED_CAST")
```
LouisJamesVasseur, line 72: Why is this annotation necessary?

LouisJamesVasseur (approved):
I agree with both the logic and the fix. Well done!

app/src/test/java/com/github/se/cyrcle/model/user/UserViewModelTest.kt
Armetiis: Because the cast of the callbacks create a warning while everything is under control 😉

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
Armetiis: The toasts are here to provide feedback when the UI is not enough. However, here the heart filling and turning red is enough for the user to understand the parking was successfully added to their favorites

sean-prz (approved):
Thanks for the fix ! This is a better and cleaner way to get to the same result. 
And many thanks for updating the documentation graph, much easier to understand than reading callbacks...

PR #294, opened 2024-11-24
Interactions with rectangles

### LINKED ISSUE #251: Rectangles aren't clickable

sean-prz:
- closes  #251 
### Simplifying the map screen.

This PR removes the three states used to manage the life cycle of a preview cards.
The listener on `MapIdleCallback` has been removed, removing the need to manage it, making `cancelables` and `removeViewAnnotation` useless.

The Preview card lifecycle is now easier to manage. 

```patch
-  var removeViewAnnotation = remember { true }
-  var cancelables = remember { Cancelable {} }
-  var listener: MapIdleCallback?
```

Before : On click, the boolean state is set to true, then a mapIdleCallback is created, and when it triggers, it recreate a card, then when the user move, the listener has to be canceled and the boolean set back to false.
Instead :  A Preview Card is drawn on Click and removed when a movement is detected. 
_(In depth: This simplification was made possible by moving the .`removeAllViewAnnotation` from a listener on the camera state to a listener on movement.)_ 

### Modifying the signature of DrawRectangles
Instead of taking a list of location it now takes a list of Parking, because for the preview Card we need all the Parking data.

```patch
  * Draw the rectangles on the map
  *
  * @param polygonAnnotationManager the polygon annotation manager
- * @param locationList the list of locations to draw
+ * @param parkingsList the list of locations to draw
  */
 fun drawRectangles(
     polygonAnnotationManager: PolygonAnnotationManager?,
     plabelAnnotationManager: PointAnnotationManager?,
-    locationList: List<Location>
+    parkingsList: List<Parking>
 ) {..}
```
### Copying the  marker click listener for polygon click listener
Once these changed have been made, it was just a copy/paste of the listener on click markers.

### Moving the annotationManager into the viewmodel

Finally the annotationManager has been moved in the viewmodel so that if a composable need later to launch an action on the preview Card, it can. 
This change was made to enable the My location button to remove the preview Card

#### Making the rectangles clickable match the expected behaviour, if the markers are clickable there are no reasons why rectangles should not be.

-- -
## Code coverage

- **MapViewModel**: 84%
- **MapScreen** : 77% (Listener on rectangles not tested, like the one on marker click :/ )

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
@@ -383,9 +383,11 @@ fun MapScreen(
       ZoomControls(
           modifier = Modifier.align(Alignment.TopEnd),
           onZoomIn = {
+            mapViewModel.removePreviewCard()
             mapViewportState.setCameraOptions { zoom(mapViewportState.cameraState!!.zoom + 1.0) }
           },
           onZoomOut = {
+            mapViewModel.removePreviewCard()
             mapViewportState.setCameraOptions { zoom(mapViewportState.cameraState!!.zoom - 1.0) }
           })
 
@@ -399,6 +401,7 @@ fun MapScreen(
                     .scale(1.2f)
                     .testTag("recenterButton"),
             onClick = {
+              mapViewModel.removePreviewCard()
               mapViewModel.updateTrackingMode(true)
               mapViewportState.transitionToFollowPuckState(
                   FollowPuckViewportStateOptions.Builder()
--- a/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
@@ -35,7 +35,6 @@ import androidx.compose.ui.unit.dp
 import com.github.se.cyrcle.R
 import com.github.se.cyrcle.databinding.ItemCalloutViewBinding
 import com.github.se.cyrcle.model.map.MapViewModel
-import com.github.se.cyrcle.model.parking.Location
 import com.github.se.cyrcle.model.parking.Parking
 import com.github.se.cyrcle.model.parking.ParkingViewModel
 import com.github.se.cyrcle.model.user.UserViewModel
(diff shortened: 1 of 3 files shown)


matttsss (commented):
When testing the App and clicking on a rectangle in advanced mode, and navigating to the `ParkingDetailsScreen`, when I go back the mode is back to default, do you think there would be an easy fix for this that you could do? I'm guessing it would be adding the mode to the `MapViewModel`?

Pijukai:
I think adding the mode to the view model should do it since the view Model are not destroyed while changing the screen whereas map is. I don't think it is limited to the mode though. Since `viewAnnotationManager` is in the View Model, all annotation managers (Point and Rectangles) should be as well (and draw markers and rectangles too). This way the logic is really separated from the UI. This was a remark on my M1 so I prefer to point out so we can improve.

Pijukai:
> > When testing the App and clicking on a rectangle in advanced mode, and navigating to the `ParkingDetailsScreen`, when I go back the mode is back to default, do you think there would be an easy fix for this that you could do? I'm guessing it would be adding the mode to the `MapViewModel`?
> 
> That is unrelated to this PR, but yes I agree that adding the mode to the viewmodel would fix this.
> 
> @Pijukai I don't mind if you want to move more stuffs to the viewmodel.

I will take care of it with my next PR bugfix that adress a bug on the MapScreen.

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -242,101 +241,116 @@ fun MapScreen(
...
+                mapMode.value = userMapMode.value
+              }
+
+              // store the zoom level
+              // This must stay at the end of the listener.
+              zoomState.value = mapView.mapboxMap.cameraState.zoom
+            }
             // =======================  CAMERA  LISTENER  =======================
 
             onDispose {
```
matttsss, line 350: Shouldn't we call `removePreviewCard` and set the annotation manager to null in the ViewModel on dispose?

sean-prz: Yes we could, that would not save a lot of memory but let me add it anyway

Pijukai (approved):
Everything is good for me. The rework is very clean great job !

matttsss (approved):
All good, thank you for the changes!

PR #297, opened 2024-11-25
Add Personal Notes Feature for Parking Details

Armetiis:
This PR introduces a new feature that allows users to add, edit, and view personal notes for parking spots. The changes include backend support, UI implementation, and corresponding tests.

#### Backend
- Added `personalNotes` field to the `UserDetails` data class
- Implemented `editCurrentUserPersonalNoteForParking` function in `UserViewModel`

#### UI
- Updated `ParkingDetailsScreen` to display and edit personal notes
- Implemented a new row in the parking details screen for note interaction
- Added icons for adding, editing, and saving notes
- Ensured proper alignment and spacing of UI elements

image
image
image

### Implementation details of editCurrentUserPersonalNoteForParking
Unified Functionality: The function handles all three operations (add, edit, remove) based on the content of newNote:
- If newNote is blank, it removes the existing note (if any).
- If newNote is not blank, it either adds a new note or updates an existing one.

### Testing
- Added new UI tests in **ParkingDetailsScreenTest** to cover note-related functionality
- Added tests for the UserViewModel's new editCurrentUserPersonalNoteForParking function
- Updated existing tests to accommodate new changes

### Coverage
image

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/model/user/User.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/user/User.kt
@@ -1,5 +1,7 @@
 package com.github.se.cyrcle.model.user
 
+const val MAX_NOTE_LENGTH = 48
+
 /**
  * Data class representing the public information of a user.
  *
--- a/app/src/main/java/com/github/se/cyrcle/model/user/User.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/user/User.kt
@@ -32,7 +32,8 @@ data class UserDetails(
     val email: String = "",
     val favoriteParkings: List<String> = emptyList(),
     // val lastLoginTime: Timestamp? = null,
-    val wallet: Wallet = Wallet.empty()
+    val wallet: Wallet = Wallet.empty(),
+    val personalNotes: Map<String, String> = emptyMap()
 )
 
 /**
--- a/app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
@@ -122,10 +122,6 @@ class UserViewModel(
 
     Log.d("UserViewModel", "Updating user: $user")
     // First update the user object in the view model.
-    if (currentUser.value == user) {
-      Log.d("UserViewModel", "No difference between the current user and the user to update")
-      return
-    }
     setCurrentUser(user)
     // Then update the user object in the database
     if (user.localSession?.profilePictureUri.isNullOrBlank() || context == null) {
--- a/app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
@@ -122,6 +122,10 @@ class UserViewModel(
 
     Log.d("UserViewModel", "Updating user: $user")
     // First update the user object in the view model.
+    if (currentUser.value == user) {
+      Log.d("UserViewModel", "No difference between the current user and the user to update")
+      return
+    }
     setCurrentUser(user)
     // Then update the user object in the database
     if (user.localSession?.profilePictureUri.isNullOrBlank() || context == null) {
--- a/app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
@@ -287,6 +287,25 @@ class UserViewModel(
   }
 
   /**
+   * Edits the personal note of the current user for the given parking. If the new note is blank,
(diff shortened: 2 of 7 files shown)


app/src/main/java/com/github/se/cyrcle/model/user/User.kt
```diff
@@ -1,5 +1,7 @@
 package com.github.se.cyrcle.model.user
 
+const val MAX_NOTE_LENGTH = 48
```
LouisJamesVasseur, line 3: Seems a little short to me. Non-blocking of course

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
```diff
@@ -122,6 +122,10 @@ class UserViewModel(
 
     Log.d("UserViewModel", "Updating user: $user")
     // First update the user object in the view model.
+    if (currentUser.value == user) {
+      Log.d("UserViewModel", "No difference between the current user and the user to update")
```
sean-prz, line 126: The wallet is not immutable.
If the user number of coin changes and only that, I believe this condition will evaluate to true and the update be skipped. 

I think we need to redefine the equal method of the user class if we want to check for equality

app/src/main/res/values/strings.xml
```diff
@@ -101,6 +101,8 @@
     <string name="list_screen_rating">Rating: %.1f</string>
     <string name="list_screen_display_only_cctv">Only display parkings with CCTV camera</string>
     <string name="all_parkings_radius">All parkings in a radius of %1$d m</string>
+    <string name="note">Edit your note</string>
```
sean-prz, line 104: These strings does not respect the convention : ScreenName_Variable

app/src/main/java/com/github/se/cyrcle/model/user/User.kt
Armetiis: Mhhh, I conceived note to be an "Alternative title", one that is more familiar to the user. This is way I chose to make it as long as the title

LouisJamesVasseur:
> > ur feature works for signed-out users should be changed: in our app thus far, we've clearly defined actions that signed-out users can and cannot do and this was reflected in the UI. Here, you still let signed out users add the note but it immediately disappears when entered. I think the option should be disabled from the get-go (the signed-out user shouldn't be able to click on the "+" button) for consistency reasons w/ the rest of the app.
> 
> You are completely right. I will make a fix for that: Adding a toast sounds appropriate to me. I should also change the text. Does this sound right to you?

Absolutely! Perfect fix in my opinion :) ! Please re-request my review once you're done!

sean-prz:
image
image
This is more of a design choice but I prefer the second version. 
What are your thoughts @LouisJamesVasseur @Armetiis

LouisJamesVasseur:
> image image This is more of a design choice but I prefer the second version. What are your thoughts @LouisJamesVasseur @Armetiis

Either works for me. Italics is a good call

Armetiis:
I also took the liberty to not display the text prompting the user to add a note if as guest. Here is what happens when clicking on the button
image

app/src/main/res/values/strings.xml
Armetiis: Fixed

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
```diff
@@ -122,6 +122,10 @@ class UserViewModel(
 
     Log.d("UserViewModel", "Updating user: $user")
     // First update the user object in the view model.
+    if (currentUser.value == user) {
```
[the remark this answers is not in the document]
Armetiis: Well the fact is that if the user saves a note without editing it, it would update the user and therefore an unnecessary write to the database

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
sean-prz: I think you can  remove the optimi. 
The update user is a very small fraction of all our API calls

Armetiis: Done :(

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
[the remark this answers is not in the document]
Armetiis: Update: removed the optimization following Sean's remark on the mutability of the wallet

sean-prz (approved):
Nice, excited to showcase this at the next meeting

LouisJamesVasseur (approved):
Perfect! Thanks for addressing my comment!

PR #303, opened 2024-11-26
Make Image clickable to expand 

sean-prz:
#### This small PR adds a click interaction on the image to expend them in an alert dialog on the detail screen.

image

### Adding a second alert dialog on the screen
- A second alert dialog has been added.
- The first alert dialog had to be renamed to be more specific:


app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsAlertDialogShowImage.kt.kt
```diff
@@ -0,0 +1,45 @@
...
+      onDismissRequest = onDismiss,
+      content = {
+        Box(
+            modifier =
+                Modifier.fillMaxWidth()
+                    .background(MaterialTheme.colorScheme.background)
+                    .padding(16.dp)
+                    .wrapContentSize()) {
+              Image(
+                  painter = rememberAsyncImagePainter(imageUrl),
```
LouisJamesVasseur, line 40: As a sanity check, what happens here if the imageUrl is invalid?

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
```diff
@@ -315,7 +323,7 @@ fun ParkingDetailsScreen(
 
               // Images
               Row(
-                  modifier = Modifier.fillMaxWidth().testTag("ImagesRow"),
+                  modifier = Modifier.height(150.dp).fillMaxWidth().testTag("ImagesRow"),
```
LouisJamesVasseur, line 326: Static height should be changed to fit the screen

LouisJamesVasseur (changes requested):
Cool feature! 
Slight issue: when testing on images that are too big for the screen (or fit the screen exactly), there's no way to go back since there's no buttons anywhere unless I haven't found how to do it (regardless, in that case, UI should clearly show how). Test with the image currently in the app for the first Parking on the Small Phone to see what I'm talking about!

sean-prz:
@LouisJamesVasseur Good remark I will add a back arrow button on the dialog

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsAlertDialogShowImage.kt.kt
sean-prz: The URL of the dialog is the same as the URL of the image on the row of images. So if the URL is invalid, the first image can't be displayed on the row and so we won't be able to instanciate a dialog box with an invalid URL

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
sean-prz: Do you mean like what you often do where the you set the dimensions to a fraction of the viewport ?
If that is the case, I prefer keeping 150.dp, less is too small to see the image, and if I adjust the size in relation to the screen height I risk displaying the images not big enough to see anything.

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsAlertDialogShowImage.kt.kt
LouisJamesVasseur: Understood, thanks for explaining!

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
LouisJamesVasseur: Fair point

LouisJamesVasseur (approved):
Great work, thanks for addressing my requested change!

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsAlertDialogShowImage.kt.kt
```diff
@@ -0,0 +1,71 @@
...
+  BasicAlertDialog(
+      modifier =
+          Modifier.testTag("ParkingDetailsAlertDialogShowImage").padding(0.dp).wrapContentSize(),
+      onDismissRequest = onDismiss,
+      content = {
+        Box(
+            modifier =
+                Modifier.heightIn(max = alertDialogMaxHeight)
+                    .wrapContentSize()
+                    // .background(MaterialTheme.colorScheme.background, MaterialTheme.shapes.small)
```
AlexandreBeeDeeBop, line 47: What are those comments for ?

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsAlertDialogShowImage.kt.kt
```diff
@@ -0,0 +1,71 @@
...
+ * Alert dialog to show the image of a parking spot. This composable is displayed when the user
+ * clicks on an image of a parking spot.
+ *
+ * @param onDismiss Callback when the dialog is dismissed.
+ * @param imageUrl URL of the image to display.
+ */
+@Composable
+@OptIn(ExperimentalMaterial3Api::class)
+fun ParkingDetailsAlertDialogShowImage(onDismiss: () -> Unit, imageUrl: String) {
+  val alertDialogMaxHeight = LocalConfiguration.current.screenHeightDp.dp * 0.8f
```
AlexandreBeeDeeBop, line 35: minor detail but you can create a val for LocalConfiguration.current.screenHeightDp.dp instead of calling it twice.

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsAlertDialogShowImage.kt.kt
sean-prz: oops

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
```diff
@@ -330,14 +338,17 @@ fun ParkingDetailsScreen(
...
+                              Modifier.weight(2f).fillMaxHeight().testTag("ParkingImagesRow"),
                           horizontalArrangement = Arrangement.spacedBy(8.dp)) {
                             items(imagesUrls) { url ->
                               Image(
                                   painter = rememberAsyncImagePainter(url),
                                   contentDescription = "Parking Image",
-                                  modifier = Modifier.size(100.dp),
-                              )
+                                  modifier =
+                                      Modifier.aspectRatio(1.0F).fillMaxHeight().clickable {
```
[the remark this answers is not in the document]
sean-prz: Sure, let me try

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsAlertDialogShowImage.kt.kt
sean-prz: Yes,  I think I will just remove the second time I called it because it is useless.

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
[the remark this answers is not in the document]
sean-prz: I like it, it crop the image so that it always fit the size given, I needed to specify then that the maxwidth is 150.dp so that we get square images : 

Before/After

PR #309, opened 2024-11-28
Search Bar and Geocoding

Pijukai:
## Content of the PR

This PR implements the following elements: 

- A search bar to look for a location
- A suggestion menu to display in a UI-friendly manner the suggestions for the location 
- A settings menu to clean the UI of the map and lay the groundwork for the filters
- Geocoding 

## Suggestion Menu and Search Bar

image 
image

- The Search bar is a `OutlinedTextField` that I unfortunately could't take from the `Atom` since it is a little more custom to the map need. 
- The suggestion menu is a `ModalBottomSheet` that is displayed when the user clicks on the button (at the bottom right) that you can see above on the keyboard that was created specially for the occasion.

## Settings Menu

image

The settings was created to simplify the UI and allow it to be more scalable with regards to future parameters we will want to add like the filters. It is a basic on click on the button, the UI displays a card with the list of the different options. Click again the menu will close. 

## Geocoding

The groundwork for the geocoding was already laid down in previous PR's but this PR add the following : 

- In `AddressViewModel`, I added the possibility to make a query in the form of a string and not a point. 
- In `AddressRepositoryNominatim`, I modified the `addressFromBody()` function so that instead of deserializing only one address, the function deserialize and return all the address suggested by Nominatim API.
- In `Address` data class, I added the fields `lat` and `lon` because it makes the implementation of the geocoding easier and they are legitimate fields in an Adress.   

## Code Coverage
image

Map is at 74% cc
The Address package containing `AddressViewModel` and `AddressRepositoryNominatim` is at 93% cc

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
@@ -556,6 +556,14 @@ fun MapScreen(
       }
 }
 
+/**
+ * Composable function to display the suggestion menu.
+ *
+ * @param showSuggestions The state of the suggestion menu.
+ * @param listOfSuggestions The list of suggestions extracted from the NominatimAPI.
+ * @param mapViewportState The viewport state of the map.
+ * @param mapViewModel The ViewModel for managing map-related data and actions.
+ */
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
 fun SuggestionMenu(
@@ -610,6 +618,12 @@ fun SuggestionMenu(
       }
 }
 
+/**
+ * Composable function to display the settings menu.
+ *
+ * @param mapMode The current map mode state.
+ * @param mapViewModel The ViewModel for managing map-related data and actions.
+ */
 @Composable
 fun SettingsMenu(mapMode: State<MapViewModel.MapMode>, mapViewModel: MapViewModel) {
   Box(modifier = Modifier.fillMaxSize().padding(16.dp), contentAlignment = Alignment.TopEnd) {


app/src/androidTest/java/com/github/se/cyrcle/di/mocks/MockAddressRepository.kt
```diff
@@ -16,11 +16,17 @@ class MockAddressRepository @Inject constructor() : AddressRepository {
           postcode = "Mock Postcode",
           country = "Mock Country")
 
-  override fun search(query: String, onSuccess: (Address) -> Unit, onFailure: (Exception) -> Unit) {
+  private val mockAddress2 = mockAddress.copy(city = "Mock City 2")
```
LouisJamesVasseur, line 19: Why did you add another MockAddress here? Does the code not function if the list only has one element in it?

app/src/androidTest/java/com/github/se/cyrcle/ui/map/MapScreenTest.kt
```diff
@@ -73,45 +80,30 @@ class MapScreenTest {
...
-  }
-
-  /**
-   * Test to verify that the internal state of the zoom level decreases correctly.
-   *
-   * This test sets the content to the `MapScreen` with an initial zoom level state, performs zoom
-   * out actions, and asserts that the state value matches the minimum zoom level.
-   */
-  @Test
-  fun testInternalStateOut() {
```
LouisJamesVasseur, line 94: So all these removals come with the removal of the zoom level right?

app/src/androidTest/java/com/github/se/cyrcle/ui/map/MapScreenTest.kt
```diff
@@ -73,45 +80,30 @@ class MapScreenTest {
...
   @Test
   fun testAddParkingRules() {
     composeTestRule.setContent {
-      MapScreen(mockNavigation, parkingViewModel, userViewModel, mapViewModel, permissionHandler)
+      MapScreen(
+          mockNavigation,
+          parkingViewModel,
+          userViewModel,
+          mapViewModel,
+          permissionHandler,
```
LouisJamesVasseur, line 105: Nitpick, but putting the permission handler in the middle of the ViewModels may make things more confusing. The way it was ordered before was more intuitive

app/src/main/java/com/github/se/cyrcle/model/address/AddressRepositoryNominatim.kt
```diff
@@ -37,24 +41,39 @@ class AddressRepositoryNominatim @Inject constructor(private val client: OkHttpC
...
+      return listOf(Address())
     }
-    val jsonObject = jsonArray.getJSONObject(0)
-    val address = jsonObject.getJSONObject("address")
-    return Gson().fromJson(address.toString(), Address::class.java)
+    val addressList = mutableListOf<Address>()
+    for (i in 0 until jsonArray.length()) {
+      if (jsonArray.getJSONObject(i) != null) {
+        val jsonObject = jsonArray.getJSONObject(i)
+        val address = jsonObject.getJSONObject("address")
```
LouisJamesVasseur, line 68: This could have also been handled more cleanly with a custom adapter for the function, in the way it's been done in some of the ViewModels. Non-blocking but this is worth looking into...

app/src/androidTest/java/com/github/se/cyrcle/di/mocks/MockAddressRepository.kt
Pijukai: The idea here is to test the feature `SuggestionMenu`. This menu triggers only when the number of suggestion is superior to 1. If it is 1 suggestion only then the camera is automatically recentered to this location.

app/src/test/java/com/github/se/cyrcle/model/address/AddressRepositoryNominatimTest.kt
```diff
@@ -139,7 +140,7 @@ class AddressRepositoryNominatimTest {
     setupMockResponse(mockResponse)
 
     var resultAddress: Address? = null
-    addressRepositoryNominatim.search("0.0,0.0", { address -> resultAddress = address }, {})
+    addressRepositoryNominatim.search("0.0,0.0", { address -> resultAddress = address.first() }, {})
```
LouisJamesVasseur, line 143: I think it makes sense, but I'd still like to verify, but why is the first always chosen?

app/src/androidTest/java/com/github/se/cyrcle/ui/map/MapScreenTest.kt
```diff
@@ -73,45 +80,30 @@ class MapScreenTest {
...
-          mockNavigation, parkingViewModel, userViewModel, mapViewModel, permissionHandler, state)
-    }
-
-    for (i in 0..(defaultZoom - minZoom).toInt()) {
-      composeTestRule.onNodeWithTag("ZoomControlsOut").performClick()
-    }
-
-    // Assert the state value
-    assert(state.value == minZoom)
+    composeTestRule.onNodeWithTag("SearchBar").assertIsDisplayed()
```
andrepeiry, line 94: I think this test can go a little further, such as checking if the search bar is clickable/can be feed some text.

LouisJamesVasseur (commented):
A great (and very necessary) feature! You did a lot of work and it shows. I left a few comments that aren't necessarily super important but just questions I got when diving deeper into the code

app/src/androidTest/java/com/github/se/cyrcle/ui/map/MapScreenTest.kt
Pijukai: Yes because i removed the `ZoomControls` that weren't supposed to stay here permanently. I used the rework of the UI to remove them and unfortunately i removed the tests that were linked which was necessary. I tried to compensate by adding more relevant tests on the new features to keep the code coverage under control .

app/src/androidTest/java/com/github/se/cyrcle/di/mocks/MockAddressRepository.kt
LouisJamesVasseur: Ahh yes, indeed, makes sense!

app/src/androidTest/java/com/github/se/cyrcle/ui/map/MapScreenTest.kt
```diff
@@ -186,12 +160,94 @@ class MapScreenTest {
...
+          mockNavigation,
+          parkingViewModel,
+          userViewModel,
+          mapViewModel,
+          permissionHandler,
+          addressViewModel)
+    }
+
+    composeTestRule.onNodeWithTag("SettingsMenuButton").assertIsDisplayed()
+    composeTestRule.onNodeWithTag("SettingsMenuButton").performClick()
```
andrepeiry, line 195: I think that it is common practice to call `isClickable` before calling `performClick`, as this ensure that there is some action do perform.

app/src/androidTest/java/com/github/se/cyrcle/ui/map/MapScreenTest.kt
Pijukai: I can change it but I would like to have @andrepeiry input on this as well if possible.

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -128,6 +162,27 @@ fun MapScreen(
...
+  val listOfSuggestions = addressViewModel.addressList.collectAsState()
+
+  // Show suggestions screen
+  val showSuggestions = remember { mutableStateOf(false) }
+
+  // Chosen location by the user
+  val chosenLocation = addressViewModel.address.collectAsState()
+
+  //  // Get the virtual keyboard controller
+  //  val virtualKeyboard : SoftwareKeyboardController = LocalSoftwareKeyboardController.current
```
andrepeiry, line 184: Is there any reason to keep this ?

app/src/main/java/com/github/se/cyrcle/model/address/AddressRepositoryNominatim.kt
Pijukai: Yes but I took the already existing function and just added a for loop. Moreover it is coded this way inside the bootcamp. However, I can open an issue and make this a PR to rework the serialization and deserialization if you want ?

app/src/test/java/com/github/se/cyrcle/model/address/AddressRepositoryNominatimTest.kt
Pijukai: Because when asking for a specific point, Nominatim will answer you with only one adress hence why the list has 1 element and I get it by using `first()`.

app/src/main/java/com/github/se/cyrcle/ui/theme/Color.kt
andrepeiry, line 1: Thank you for making this available for later if we ever need it ^^

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Pijukai: Totally forgot this good catch.

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
andrepeiry, line 1: I know the MapScreen is quite Monolithic from previous interations, but could you move (at least) your parts into dedicated `@Composable` function, so that you don't need to add `// === THIS PART ===` comments ?

app/src/main/java/com/github/se/cyrcle/model/address/AddressRepositoryNominatim.kt
Pijukai: You suggests I should create an adapter or modify an already existing one ?

LouisJamesVasseur: There's good references inside of our ViewModels if you want to have a look!

app/src/androidTest/java/com/github/se/cyrcle/ui/map/MapScreenTest.kt
Pijukai: Done thanks !

andrepeiry (commented):
Potent PR, code is quite readable even though some small modularity issues.

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Pijukai: Do you think it is useful since I am using those functions only once ? `// ===== THIS PART=====` is meant to make the code clearer is the first place not compensate for anything. Tell me if it seems useful to make composable functions out of those and i will change this !

LouisJamesVasseur: I personally don't see an issue. I'm of the opinion that having a ton of @Composable functions floating around isn't always super necessary and makes it harder to find issues when re-reading it. I think @Pijukai's way of handling modularization in this function makes things quite clear, though I do see where you're coming from

andrepeiry: I think it is a good think to make them into dedicated `@Composable`. I know you only use them here, but it helps the readability of the thing. The "NASA way of thing" is to have no function longer than 60 lines of codes (pertinence may be a bit lacking, as we're doing UI, but I hope you get the idea)
The end goal is to keep in mind "One function does one thing", and having 
```
Scaffold(
   topAppBar = TopAppBar()
   bottomAppBar = BottomAppBar()
) {
   Composable1
   Composable2
   Composable3
}
```
to keep the Scaffold "as small as possible/needed", and being able to tweak each composable effectively and rapidely

app/src/main/java/com/github/se/cyrcle/model/address/AddressRepositoryNominatim.kt
Pijukai: Now that I tried putting an Adapter i remember why I couldn't. The `Address` data class was based on the fields that constitutes the `address` part of nominatim answer you can see. I did this way because i needed to extract the `lat` and ` lon` fields that are not part of the `address` field but rather parallel field. Does it seem clear to you ? If you have a solution to it I would gladly implement it but remember also that if the solution implies that we need to rework all the serialization/ deserialization it should be a part of a PR since this amount to a lot of work :+1:

LouisJamesVasseur: I agree with you on this. Let's make an issue and re-discuss this at another time. It's outside the scope of your PR as is

app/src/androidTest/java/com/github/se/cyrcle/ui/map/MapScreenTest.kt
andrepeiry: It is indeed something that would be nice to have "in-order", but it is not important to my eyes, as this is only done once at screen instantiation, so the maintenance impact isn't that big to my eyes, I feel like we can leave this as it is.

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Pijukai: Okay i understand what you mean. I refactored the `SettingsMenu` and `SuggestionMenu` to make this more clear. For the rest , it starts to be difficult due to the fact that the search bar for example is really intertwined with the MapScreen function. Is it okay for you ? You can take a look i just pushed the code.

andrepeiry (approved):
Thank you for the refactor. I understand that some things are hardly modularizable, and won't be too much picky on this. Nice work ^^

LouisJamesVasseur (commented):
Great work on those fixes :). Great PR overall!

LouisJamesVasseur (approved):

app/src/androidTest/java/com/github/se/cyrcle/ui/map/MapScreenTest.kt
LouisJamesVasseur: Fine by me too then

Armetiis:
Hi, just passing by to share some ideas and opinions for future work.
Great work you did here. Settings overlay box is a nice idea but in my opinion it should definitely be converted to a Dialog Box in a future PR. Probably the one that will add filters to the map
I would also love to investigate the possibility of sorting the outputs by relevance, in particular to the center of the camera. So if for example I am have my camera over Lausanne and look for Avenue de La Gare, it suggests the ones that are the most relevant to Lausanne.
Obviously this is clearly out of the scope of this PR, which is very complete.

Pijukai:
> Hi, just passing by to share some ideas and opinions for future work. Great work you did here. Settings overlay box is a nice idea but in my opinion it should definitely be converted to a Dialog Box in a future PR. Probably the one that will add filters to the map I would also love to investigate the possibility of sorting the outputs by relevance, in particular to the center of the camera. So if for example I am have my camera over Lausanne and look for Avenue de La Gare, it suggests the ones that are the most relevant to Lausanne. Obviously this is clearly out of the scope of this PR, which is very complete.

For the relevance of the suggestions i think it is a great idea, let's see how it affect performance but definitely something to keep in mind i agree. For the Dialog Box you may be right, i just added the settings menu to integrate correctly the search bar. It can be changed and  I just implemented the easiest thing that would give the best  functional result without implementing advanced graphics.  But some good ideas to keep in mind thanks :+1:

PR #325, opened 2024-11-29
UI rework of All Reviews screen

### LINKED ISSUE #323: On All Reviews screen, clicking report as guest makes app crash

### LINKED ISSUE #324: Rework all reviews screen
- Rescale the card to wrap its content
- Replace report button by dropdown menu option
- Pin current user's review on top
- Also replace edit and delete buttons by dropdown menu options

Armetiis:
This pull request introduces UI improvements to the review card UI in the application. The changes include a complete redesign of the review card layout, the introduction of a dropdown menu for options instead of floating buttons, and modularization of the ReviewCard into its own composable, aswell as rescaling of the cards.

image
image
image

### Changes in the overrall layout of the screen
- **Distinct Sections:** Introduced separate sections for the current user's review and other users' reviews. The first section is only visible to signed-in users. If the user has not submitted a review, this section contains an "Add Review" button.
- **Dropdown Menu for Options:** Replaced floating buttons for editing and deleting a review with a dropdown options menu that is visible only in the current user's card.

### Changes in the card
- **Rescaled Card:** The review card has been rescaled to better fit its content, displaying two lines of text when not expanded, with ellipsis for overflow. Clicking on the card expands it to show full content.
- **Color Rework:** Adjusted colors to use a less aggressive surface color for the card, combining black and primary colors for the content.
- **Layout Adjustments:** Modified the layout of elements within the card; now, the score and date are displayed on the same line.
- **Dropdown Menu Integration:** Replaced the report button with an option in the dropdown menu.

### Implementation changes
- **New Options Menu Composable:** Added a new composable for the options menu that accepts a map of option names and their corresponding callbacks.
- **Star Rating Text Scaling:** Scaled text in the stars composable according to a provided scale factor.

### Fixes:
Fixed an issue that prevented signed-out users from reporting reviews, which previously caused app crashes. Closes #323.

### Coverage
image

-- -
Closes #324

### DIFF
--- a/app/src/androidTest/java/com/github/se/cyrcle/ui/review/AllReviewsScreenTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/ui/review/AllReviewsScreenTest.kt
@@ -88,8 +88,18 @@ class AllReviewsScreenTest {
     composeTestRule.onNodeWithTag("ReviewCard0").performClick()
     composeTestRule.onNodeWithText("New Review.").assertIsDisplayed()
 
-    composeTestRule.onNodeWithTag("YourReviewTitle").assertIsDisplayed()
+    composeTestRule.onNodeWithTag("YourReviewTitle").assertIsNotDisplayed()
     composeTestRule.onNodeWithTag("OtherReviewsTitle").assertIsDisplayed()
+  }
+
+  @Test
+  fun signedInUser_seesTwoSectionTitles() {
+    composeTestRule.setContent {
+      AllReviewsScreen(navigationActions, parkingViewModel, reviewViewModel, userViewModel)
+      userViewModel.setCurrentUser(TestInstancesUser.user1)
+    }
+
+    composeTestRule.onNodeWithTag("YourReviewTitle").assertIsDisplayed()
   }
 
   @Test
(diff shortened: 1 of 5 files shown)


app/src/androidTest/java/com/github/se/cyrcle/ui/review/AllReviewsScreenTest.kt
LouisJamesVasseur, line 1: A thought: since we're now highlighting the user's review, shouldn't selecting the User be a part of the setUp for this test suite? Before the UI only checked whether a review was added when the Edit/Delete buttons are but now it checks it straight away when the screen is set. When I tried running the third test in isolation for instance, I got an error because the user wasn't set. What do you think?

LouisJamesVasseur (approved):
I have looked as deeply as I can and can't think of something to say, everything makes sense. You really did an excellent job and made the AllReviewsScreen I wish I had back in Sprint 4. If I had to say something, I think the animations when you clicked on a review card were quite nice and I'd like to see them back at some point!

sean-prz (approved):
Wow, lovely, thanks for always making our app feels more professional !

PR #326, opened 2024-11-29
Add check for incorect fields during account creation

### LINKED ISSUE #295: CreateProfileScreen : Inputs field are not validated
### Problem
 a user can create a profile with invalid input field, i.e an empty username.

The UI correctly shows the invalid input, but the logic still accepts it

matttsss:
## What is this PR?
This PR adds checks for invalid fields during the creation/editing of the account.
It also adds the package name to the `GamblingScreenTest` since it seems that without that, those tests were not run.

Closes #295 

### Test Coverage:
image


sean-prz (approved):
The bugfix works, nice catch on the required `.setCurrentUser() `
All is OK

Armetiis (approved):
Good job. I tested it with my 3rd gmail address and it worked as expected. Tried making sure that the base of the authentication still worked, but given how solid it is I doubted it would fail.
Nothing to add 👍

PR #327, opened 2024-11-29
Preliminary setup for Zone Management. 

sean-prz:
- closes : #328 
- closes : #329 
#### This PR includes the preliminary changes to make the zone management, necessary for our offline mode.
The only notable change for the app is on the settings popup : 
image

## How
-  `Zone` Data class (unused for now) 
A zone is an area of the map the user has selected for download.
- `ZoneManagerScreen`, the screen from which a user will be able to manage all of his zones stored on his deviced.
- `ZoneSelectionScreen` the screen accessible from `ZoneManagerScreen` to allow user to create new zone.
-  new Row on the settings of mapScreen to access the `ZoneManagerScreen`.

_I will not submit this PR for grading as it does not bring any functionality, this PR is all the boilerplate, boring code so that the next ones can be centered around functionnality_ 

-- -

## Code coverage
ui.zone : 74%
model.zone : 0% 
(Obviously as this is only boilerplate code and the zones are not used the cc is really low)

image how are we at 80%, can anybody check?

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/zone/ZoneManagerScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/zone/ZoneManagerScreen.kt
@@ -1,0 +1,11 @@
+package com.github.se.cyrcle.ui.zone
+
+import androidx.compose.runtime.Composable
+import com.github.se.cyrcle.ui.navigation.NavigationActions
+import com.github.se.cyrcle.ui.theme.atoms.Text
+
+/** Screen where users can manage their downloaded zones. and access the zone selection screen. */
+@Composable
+fun ZoneManagerScreen(navigationActions: NavigationActions) {
+  Text("Zone Manager")
+}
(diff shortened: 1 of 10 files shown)


matttsss (approved):
All good for me. Not much to say except that we have a long road in front of ourselves 🫡

app/src/main/java/com/github/se/cyrcle/ui/zone/ZoneManagerScreen.kt
```diff
@@ -0,0 +1,53 @@
...
+@Composable
+fun ZoneManagerScreen(navigationActions: NavigationActions) {
+  Scaffold(
+      topBar = {
+        TopAppBar(
+            title = stringResource(R.string.zone_manager_screen_title),
+            navigationActions = navigationActions)
+      }) { padding ->
+        Box(modifier = Modifier.padding(padding).fillMaxSize()) {
+          AddZoneButton(navigationActions)
```
Armetiis, line 33: It's not that important, but here it would make it much easier if you just used the "floating action button" parameter of the scaffold rather than doing it like this. You wouldn't even have to worry about using a box and paddings and stuff.

Armetiis (approved):
I don't have much to add. This is pretty straighforward. For now, this is understandable xD

Issue #317, opened 2024-11-29
Add a Zone management screen.


sean-prz:

sean-prz:
Add a screen where a user can manage all of the zone he downloaded. 
Accessible from the Map Screen.

PR #332, opened 2024-11-30
Add and Manage zones

sean-prz:
- fix:  #317 
- fix: #335 
#### Users can now manage (add/refresh/delete) offline portions of the map called Zone. 
The informations about the zones, is downloaded on the local file storage, meaning the user can access this data, no matter what account, he's logged in, even as guest mode.

image
image

## How
- The `ZoneManagerScreen` load from files all zones and displays it, with possibility to refresh or delete it.
- The `ZoneSelectionScreen` allow user to add a new zone with code copied from `LocationPicker` and a dialog box to pick the name.
- The `ZoneSelectionScreen` write to the local storage the information about the zone.
- All the IOs are done in the `Zone.kt` file. Ensurint that the path and naming is standarised
- The `Zone` Object are stored as json object in` appDir/zones/${zone.uid}.json`

**Important** Currently when adding a zone, the tile and parking data is not downloaded, same for deletion and refresh. => Only the Zone Data Class is managed not the Parkings and not the Map. 
(Once we have implemented the function we should call them in the callbacks of the button, that add/refresh/delete.)

#### This is one of the first step in order to have an offline mode. We now have the interface to manage our offline plans. 

-- -
Overall: 80%
ui zone : 75%
model zone : 25%
image

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
@@ -31,6 +31,8 @@ import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.map
 
+const val DEFAULT_ZOOM_FOR_ZONE_SELECTION = 10.0
+
 class MapViewModel : ViewModel() {
 
   private val _cameraPosition = MutableStateFlow<CameraState?>(MapConfig.defaultCameraState())
@@ -142,10 +144,19 @@ class MapViewModel : ViewModel() {
     _cameraPosition.value = newCameraState
   }
 
-  fun unZoomActualCamera() {
+  /**
+   * Unzooms the camera, to show a large are of the map to the user selecting a zone to download
+   * offline.
+   */
+  fun setCameraForZoneSelection() {
     val actualCamera = _cameraPosition.value!!
     _cameraPosition.value =
-        CameraState(actualCamera.center, EdgeInsets(0.0, 0.0, 0.0, 0.0), 10.0, 0.0, 0.0)
+        CameraState(
+            actualCamera.center,
+            EdgeInsets(0.0, 0.0, 0.0, 0.0),
+            DEFAULT_ZOOM_FOR_ZONE_SELECTION,
+            0.0,
+            0.0)
   }
 
   /**
--- a/app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
@@ -142,6 +142,12 @@ class MapViewModel : ViewModel() {
     _cameraPosition.value = newCameraState
   }
 
+  fun unZoomActualCamera() {
+    val actualCamera = _cameraPosition.value!!
+    _cameraPosition.value =
+        CameraState(actualCamera.center, EdgeInsets(0.0, 0.0, 0.0, 0.0), 10.0, 0.0, 0.0)
+  }
+
   /**
    * Update the screen coordinates states
    *
--- a/app/src/main/java/com/github/se/cyrcle/model/zone/Zone.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/zone/Zone.kt
@@ -1,34 +1,35 @@
 package com.github.se.cyrcle.model.zone
 
+import android.content.Context
 import com.mapbox.geojson.BoundingBox
-import org.json.JSONObject
 import java.io.File
-import java.time.LocalTime
+import java.time.LocalDate
 import java.util.UUID
+import org.json.JSONObject
+
+const val ZONE_DIR = "zones"
 
 data class Zone(
     val boundingBox: BoundingBox,
     val name: String,
-    val lastRefreshed: LocalTime,
+    val lastRefreshed: LocalDate,
     val uid: String
 ) {
-    /**
-     * Zone constructor that set the time of the last refresh to the current time, and generates a new
-     * UID.
-     *
-     * @param boundingBox The bounding box of the zone.
-     * @name The name of the zone chosen by the user.
-     */
-    constructor(
-        boundingBox: BoundingBox,
-        name: String
-    ) : this(boundingBox, name, LocalTime.now(), generateZoneUid())
+  /**
+   * Zone constructor that set the time of the last refresh to the current time, and generates a new
+   * UID.
+   *
+   * @param boundingBox The bounding box of the zone.
+   * @name The name of the zone chosen by the user.
+   */
+  constructor(
+      boundingBox: BoundingBox,
+      name: String
+  ) : this(boundingBox, name, LocalDate.now(), generateZoneUid())
 
-    /**
-     * Returns a JSON representation of the zone.
-     */
-    fun toJson(): String {
-        return """
+  /** Returns a JSON representation of the zone. */
+  fun toJson(): String {
+    return """
             {
                 "boundingBox": {
                     "south": ${boundingBox.south()},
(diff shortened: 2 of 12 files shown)


app/src/main/java/com/github/se/cyrcle/ui/zone/ZoneManagerScreen.kt
```diff
@@ -23,19 +45,50 @@ import com.github.se.cyrcle.ui.theme.molecules.TopAppBar
...
             navigationActions = navigationActions)
       }) { padding ->
         Box(modifier = Modifier.padding(padding).fillMaxSize()) {
+          // === Overlay ===
           AddZoneButton(navigationActions)
-          Text("TODO")
+          // ===  === ===  ===
+
+          // === Content ===
+          Column(Modifier.fillMaxWidth()) {
```
matttsss, line 67: Here I think a lazycolumn might be a better idea? We do not know the number of zones we have at compile time...

app/src/main/java/com/github/se/cyrcle/ui/zone/ZoneManagerScreen.kt
```diff
@@ -51,3 +104,57 @@ fun AddZoneButton(navigationActions: NavigationActions) {
         testTag = "AddButton")
   }
 }
+// Display relevant information about a zone. Allows the user to refresh the zone or delete it.
+@Composable
+fun ZoneCard(zone: Zone, zones: MutableState<List<Zone>>) {
```
matttsss, line 109: I guess that passing `zones` is ok, but wouldn't it be better to make it a `onDelete` callback, this might be more versatile?

app/src/main/java/com/github/se/cyrcle/ui/zone/ZoneSelectionScreen.kt
```diff
@@ -1,26 +1,226 @@
...
+) {
+  val showAlertDialogPickName = remember { mutableStateOf(false) }
+  val boundingBox = remember { mutableStateOf<BoundingBox?>(null) }
+  val zoneName = remember { mutableStateOf("") }
+  val mapView = remember { mutableStateOf<MapView?>(null) }
+  val locationPickerState by
+      mapViewModel.locationPickerState
+          .collectAsState() // state representing where the user is in the location selection
+  // process
+  mapViewModel
```
matttsss, line 64: Would it not be better to move this function call into the `LaunchedEffect(Unit)` ? I'm not to familiar with this so I'm not sure if that would be better?

matttsss (changes requested):
Overall a nice improvement... However I have a few issues with the UI:
* There are no size constraints on the name, is that a good idea?
* I also have a crash with nullptr exception, to reproduce:
    - Add zone
    - Try to add a second one
    (I have DM'ed you the crash info)

Could you try to fix this crash before we merge this PR?

sean-prz:
@matttsss 
I replaced the input text with the conditional one made by Maria (which I tweaked a bit) 
image
And fixed the crash.
The locationPickerState should have been reset in the `LaunchEffect(Unit) {...} `but somehow the `LaunchedEffect(locationPickerState) {...}` runs before and so made the app crash

app/src/main/java/com/github/se/cyrcle/ui/zone/ZoneManagerScreen.kt
sean-prz: Sure I can do a lazyColumn, did not think users would add more than 2 or 3 zones, but it's better

app/src/main/java/com/github/se/cyrcle/ui/zone/ZoneSelectionScreen.kt
sean-prz: It doesn't work, because it launch a new coroutine and the map is created before the camera is updated, making the call useless

app/src/main/java/com/github/se/cyrcle/ui/zone/ZoneManagerScreen.kt
sean-prz: I don't like callback, I think they make the code harder to read 🙈 , especially here we don't need versatility as we would never call the `ZoneCard` with a different onDelete function

matttsss: Honestly, I can understand... (specially when we look at the ViewModels xD)

matttsss (approved):
Nice, thank you for solving this issue!

app/src/main/java/com/github/se/cyrcle/model/zone/Zone.kt
```diff
@@ -20,7 +25,101 @@ data class Zone(
   constructor(
       boundingBox: BoundingBox,
       name: String
-  ) : this(boundingBox, name, LocalTime.now(), generateZoneUid())
+  ) : this(boundingBox, name, LocalDate.now(), generateZoneUid())
+
+  /** Returns a JSON representation of the zone. */
+  fun toJson(): String {
```
Armetiis, line 31: I feel like this function needs a lot of maintenance. As soon as you add a field you have to modify the toJson and fromJson functions. Wouldn't it be better to proceed like in ParkingRepositoryFirestore and have an adapter for the Bounding Box. Have a look at `BoundingBoxTypeAdapter()` provided by mapbox first to make sure it doesn't work how you'd like it to before looking into making a custom one.

If you prefer keeping a man made serializer, maybe you could use a json Builder as I really feel like this method is error prone

app/src/main/java/com/github/se/cyrcle/model/zone/Zone.kt
```diff
@@ -20,7 +25,101 @@ data class Zone(
...
+     *
+     * @param zone The zone to store.
+     * @param context The context of the application, necessary to access the file system.
+     */
+    fun storeZone(zone: Zone, context: Context) {
+      val zoneDir = File(context.filesDir, ZONE_DIR)
+      if (!zoneDir.exists()) {
+        zoneDir.mkdirs()
+      }
+      File(zoneDir, "${zone.uid}.json").writeText(zone.toJson())
```
Armetiis, line 83: beautiful

app/src/main/java/com/github/se/cyrcle/ui/addParking/location/LocationPickerBottomBar.kt
Armetiis, line 1: I am a bit confused by the 138 lines changed. I guess this is just a ktfmt thing, but I couldn't identify the actual change. Plus you don't even use in your implementation in the end. So I guess it's nothing too important

app/src/main/java/com/github/se/cyrcle/ui/theme/atoms/Input.kt
Armetiis, line 1: I think it's nice that you improve the atom, but make sure that you don't bring down the coverage. Adding branches in UI elements is the fastest way to divide coverage by 2

app/src/main/java/com/github/se/cyrcle/ui/zone/ZoneManagerScreen.kt
Armetiis: Honestly I also think it's a bit weird to pass the whole List to a card, but if you think that this way is better, it's ok

app/src/main/java/com/github/se/cyrcle/ui/zone/ZoneManagerScreen.kt
```diff
@@ -46,8 +102,65 @@ fun AddZoneButton(navigationActions: NavigationActions) {
...
+            style = MaterialTheme.typography.bodyMedium.copy(fontWeight = FontWeight.Bold))
+        Text(
+            text = stringResource(R.string.zone_manager_header_lastRefreshed),
+            modifier = Modifier.weight(1f).testTag("ZoneManagerHeaderLastRefreshed"),
+            style = MaterialTheme.typography.bodyMedium.copy(fontWeight = FontWeight.Bold))
+        Icon(
+            Icons.Filled.Refresh,
+            contentDescription = "Refresh",
+            modifier = Modifier.weight(0.2f).alpha(0.0f))
+        Icon(
```
Armetiis, line 161: Really smart to put invisible icon for alignment

Armetiis (approved):
Honestly very clean implementation, be it from a backend or UI perspective.
There's only the (de)serializer I am a bit worried about, but I don't think this change has to be addressed now. It can be addressed in the future, in particular if they need to evolve.

Keep up the good work. Soon we'll have a fully working offline mode 🫡

app/src/main/java/com/github/se/cyrcle/model/zone/Zone.kt
sean-prz: @Armetiis , You're right it's not pretty.
I fixed it take a look at the corresponding commit
I had to make a customTypeAdapter for the localDate, and learn how to use Gson, but I think it looks cleaner now.

app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
```diff
@@ -142,6 +142,12 @@ class MapViewModel : ViewModel() {
     _cameraPosition.value = newCameraState
   }
 
+  fun unZoomActualCamera() {
```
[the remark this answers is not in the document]
sean-prz: fixed !

Armetiis (approved):
Thank you for addressing the changes! Everything seems to be still working perfect. Good job

PR #337, opened 2024-12-02
Feature download parkings

matttsss:
## What is this PR?
This PR aims to enable the download and management of parkings in zones for offline mode.
It adds two new functions to the `ParkingViewModel`: 
* `downloadZone(...)` that downloads the parkings for a certain zone
* `deleteZone(...)`  that deletes the downloaded parkings that are within that zone

### Main changes:
* `Parking` now has a new attribute: `tile` which identifies to what tile it belongs
* `Tile` is now a type alias for `String`, this simplifies the management and creation of tiles, the "static" functions that were associated to it are now in a new object `TileUtils`.
* A new interface `OfflineParkingRepository` that extends `ParkingRepository`, it adds the API for dowloading / deleting zones
* `ParkingViewModel` has a new attribute of type `OfflineParkingRepository`

### Testing the database:
The Unit/Android tests cover the database by testing the `OfflineParkingRepositoryRoom`.
When the app is running, it is possible to inspect the database via the "App Inspector" as explained here:
https://developer.android.com/studio/inspect/database

## Code coverage:
- `com.github.se.cyrcle.model.parking.offline`: 100%
- `ParkingViewModel`: 86%
- Global:
image


app/src/androidTest/java/com/github/se/cyrcle/model/parking/OfflineParkingRepositoryRoomTest.kt
```diff
@@ -0,0 +1,230 @@
...
+            .build()
+    parkingDao = database.parkingDao
+    repository = OfflineParkingRepositoryRoom(database)
+
+    parkingDao.upsert(TestInstancesParking.parking1)
+  }
+
+  @After
+  fun tearDown() {
+    database.close()
```
andrepeiry, line 50: *[TEMPORARY COMMENT WHILE ANDROID STUDIO IS SYNCING]*
Wouldn't it be cleaner to also destroy the database as well ? I feel like this can be some bug source if we don't delete the database between tests. I'm ready to believe that JUnit recreates the whole environment each time, but I'm not an expert on this subject.

matttsss: Mhhhh I'm not to sure. I think that because the `@Before` creates a new database, it ovewrites the database already created and held in that reference. Since those test databases are in memory, they will simply be deleted by the garbage collector?

app/src/main/java/com/github/se/cyrcle/io/serializer/ParkingAdapter.kt
andrepeiry, line 1: For this file, I'm bothered with the following :
- Documentation is missing on most of the function
- Exception handling could be done at this level, as one of the two following, notably for the enum, in which the ordinal can be corrupt and bigger than the number of entries
  - change the function to the following scheme `deserializeOrDefault(ordinal, default) { try {deserialize} catch(E) {return default} }`
  - Throw a "proper" deserialization error instead of relying on the "IndexOutOfBound` when dealing with enums.

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
```diff
@@ -194,31 +189,25 @@ class ParkingViewModel(
...
-      // Get parkings from memory cache
-      if (tileCache.containsKey(tile.uid)) {
-        _rectParkings.value += tileCache[tile.uid]!!.parkings
+      //  Cache hit
+      if (tilesToParking.containsKey(tile)) {
+        _rectParkings.value += tilesToParking[tile]!!
         updateClosestParkings(--nbRequestLeft)
 
-        // Get parkings from repository
+        // Cache miss
```
andrepeiry, line 203: Just a quick note (non-blocking), is that an explanation for an `if-else` statement should be written as follow
```
if (condition) {
    // Explanation
    ... code ...
} else {
    // Explanation
    ... code ...
}
```
Instead of above the branching, as this would better indicate the whole thing
```
// Absolute value of x
if (x < 0) {
    x =-x;
} else {
    x = x;
}
```

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
```diff
@@ -659,18 +648,61 @@ class ParkingViewModel(
...
+  fun deleteZone(zoneToDelete: Zone, allZones: List<Zone>) {
+    val tilesToKeep =
+        allZones
+            .flatMap { zone ->
+              if (zone != zoneToDelete)
+                  TileUtils.getAllTilesInRectangle(
+                      zone.boundingBox.southwest(), zone.boundingBox.northeast())
+              else emptySet()
+            }
+            .toSet()
```
andrepeiry, line 699: I haven't tried myself, but could this have been obtain more easily with a `.filter` ?

app/src/main/java/com/github/se/cyrcle/model/parking/Tile.kt
```diff
@@ -1,128 +1,139 @@
...
+        Point.fromLngLat(eastPoint.longitude(), northPoint.latitude()))
+  }
+
+  /**
+   * Returns the bottom left point of the tile
+   *
+   * @param tile the tile to get the bottom left point from
+   * @return the bottom left point of the tile
+   */
+  fun getBottomLeftPoint(tile: Tile): Point? {
```
andrepeiry, line 98: The documentation for `getBottomLeftPoint` as well as `getTopRightPoint` could be improve such as
```
@param tile a valid tile to get the bottom left point from.
@return the bottom left point of the tile, null if tile is not valid.
```

app/src/main/java/com/github/se/cyrcle/model/parking/online/ParkingRepositoryFirestore.kt
```diff
@@ -68,12 +71,19 @@ class ParkingRepositoryFirestore @Inject constructor(private val db: FirebaseFir
...
-  override fun getParkingsBetween(
-      start: Point,
-      end: Point,
+  override fun getParkingsForTile(
+      tile: Tile,
       onSuccess: (List<Parking>) -> Unit,
       onFailure: (Exception) -> Unit
   ) {
+    // Get the bottom left and top right points of the tile
+    // This is a temporary solution for the time that `Parking.tile` gets added on Firestore
```
andrepeiry, line 80: Please mark this with a `TODO` or another easy searchable string so that we can find them later ^^

app/src/main/java/com/github/se/cyrcle/model/parking/online/ParkingRepository.kt
andrepeiry, line 1: Thank you for adding the documentation on `getReportsForParking` and `addReport` <3

andrepeiry (changes requested):
# Good code in general

Your code is readable, and even if the PR is quite hefty, it is not a pain to read.

## Some improvements

I'm a bit bothered by the `Adaptater`s

- Documentation is missing on most of the function
- Exception handling could be done at this level, as one of the two following, notably for the enum, in which the ordinal can be corrupt and bigger than the number of entries (also don't forget to add some tests on this)
   - change the function to the following scheme `deserializeOrDefault(ordinal, default) { try {deserialize} catch(E) {return default} }`
   - Throw a "proper" deserialization error instead of relying on the `IndexOutOfBound` when dealing with enums.

Also see some inline comments I left inside the code

app/src/androidTest/java/com/github/se/cyrcle/model/parking/OfflineParkingRepositoryRoomTest.kt
LouisJamesVasseur: Would you be willing to test both and report back? This could be a MAJOR error source so I'd be grateful if you tried before merging the PR. I am however not an expert either, so if you're confident I'll take your word for it

app/src/androidTest/java/com/github/se/cyrcle/ui/zone/ZoneManagerScreenTest.kt
```diff
@@ -9,31 +9,52 @@ import androidx.compose.ui.test.junit4.createComposeRule
...
 import com.github.se.cyrcle.ui.navigation.Screen
 import com.mapbox.geojson.BoundingBox
 import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
+import org.junit.runner.RunWith
 import org.mockito.Mockito.mock
 import org.mockito.kotlin.verify
 
+@RunWith(AndroidJUnit4::class)
```
LouisJamesVasseur, line 30: any idea why it wasn't using the runWith before?

app/src/main/java/com/github/se/cyrcle/io/serializer/LocationAdapter.kt
LouisJamesVasseur, line 1: As @andrepeiry mentions, documentation is partly missing here

app/src/main/java/com/github/se/cyrcle/io/serializer/ParkingAdapter.kt
```diff
@@ -12,37 +17,50 @@ import com.google.gson.reflect.TypeToken
...
-   */
   @TypeConverter
-  fun serialize(parking: Parking): String {
-    return gson.toJson(parking)
+  fun serializeParkingCapacity(parkingCapacity: ParkingCapacity): Int {
+    return parkingCapacity.ordinal
+  }
+
+  @TypeConverter
+  fun deserializeParkingCapacity(ordinal: Int): ParkingCapacity {
```
LouisJamesVasseur, line 42: Yeah I definitely agree w/ @andrepeiry here. If we're dealing w/ indices in a table we need to error handle... it'd be quite a shame for the app to consistenly crash if we make a change at this level

app/src/main/java/com/github/se/cyrcle/io/serializer/ParkingAdapter.kt
```diff
@@ -12,37 +17,50 @@ import com.google.gson.reflect.TypeToken
...
-   */
   @TypeConverter
-  fun deserialize(data: String): Parking {
-    return gson.fromJson(data, Parking::class.java)
+  fun serializeParkingRackType(parkingRackType: ParkingRackType): Int {
+    return parkingRackType.ordinal
+  }
+
+  @TypeConverter
+  fun deserializeParkingRackType(ordinal: Int): ParkingRackType {
```
LouisJamesVasseur, line 52: same comment as above

app/src/main/java/com/github/se/cyrcle/io/serializer/ParkingAdapter.kt
```diff
@@ -12,37 +17,50 @@ import com.google.gson.reflect.TypeToken
...
+  }
+
+  @TypeConverter
+  fun serializeParkingProtection(parkingProtection: ParkingProtection): Int {
+    return parkingProtection.ordinal
+  }
+
+  @TypeConverter
+  fun deserializeParkingProtection(ordinal: Int): ParkingProtection {
+    return ParkingProtection.entries[ordinal]
```
LouisJamesVasseur, line 63: here too :)

app/src/main/java/com/github/se/cyrcle/di/DatabaseModule.kt
```diff
@@ -0,0 +1,30 @@
...
+import dagger.Module
+import dagger.Provides
+import dagger.hilt.InstallIn
+import dagger.hilt.android.qualifiers.ApplicationContext
+import dagger.hilt.components.SingletonComponent
+import javax.inject.Singleton
+
+@Module
+@InstallIn(SingletonComponent::class)
+object DatabaseModule {
```
LouisJamesVasseur, line 15: Maybe some doc here?

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
```diff
@@ -194,31 +189,25 @@ class ParkingViewModel(
...
-        // Get parkings from repository
+        // Cache miss
       } else {
-        // Cache tile
-        tileCache[tile.uid] = tile
-        Log.d("ParkingViewModel", "Requesting parkings for tile ${tile.uid}")
-        parkingRepository.getParkingsBetween(
-            tile.bottomLeft,
-            tile.topRight,
+        tilesToParking[tile] = emptyList()
```
LouisJamesVasseur, line 205: Why is this here? This is "false" branching of the if condition so that means that tilesToParking[tile] is already an empty list, no?

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
matttsss: I guess? But since there is only one element to filter out, I'm not sure if it would be that useful...

app/src/androidTest/java/com/github/se/cyrcle/ui/zone/ZoneManagerScreenTest.kt
matttsss: No, but it can only help to have them

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
```diff
@@ -659,18 +648,61 @@ class ParkingViewModel(
...
+                  TileUtils.getAllTilesInRectangle(
+                      zone.boundingBox.southwest(), zone.boundingBox.northeast())
+              else emptySet()
+            }
+            .toSet()
+
+    val tilesToDelete =
+        TileUtils.getAllTilesInRectangle(
+            zoneToDelete.boundingBox.southwest(), zoneToDelete.boundingBox.northeast())
+    offlineParkingRepository.deleteTiles(tilesToDelete - tilesToKeep) {
```
LouisJamesVasseur, line 704: Again, a quick error handle to check that tilesToDelete > tilesToKeep would be nice

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
matttsss: Since requests to Firestore are asynchronous, if we have a second call to `getParkingsInRect`, it will make a second request to firestore if the first one did not complete before. This way the empty list marks this tile as "already" queried

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
matttsss: Since those are sets, the difference can only be the empty set in the worst case :)

app/src/main/java/com/github/se/cyrcle/model/parking/offline/OfflineParkingRepositoryRoom.kt
LouisJamesVasseur, line 1: documentation here please :)

app/src/main/java/com/github/se/cyrcle/ui/zone/ZoneSelectionScreen.kt
```diff
@@ -96,19 +100,30 @@ fun ZoneSelectionScreen(
...
-          createZone(boundingBox.value!!, it, context, navigationActions)
+          coroutineScope.launch {
+            val zone = Zone.createZone(boundingBox.value!!, it, context)
+            parkingViewModel.downloadZone(
+                zone,
+                { navigationActions.goBack() },
+                {
+                  // On failure, avoid keeping stale zone
+                  Zone.deleteZone(zone, context)
+                  Toast.makeText(context, "Error downloading zone", Toast.LENGTH_SHORT).show()
```
LouisJamesVasseur, line 121: Should be in strings.xml

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
LouisJamesVasseur: Good point!

app/src/androidTest/java/com/github/se/cyrcle/ui/zone/ZoneManagerScreenTest.kt
LouisJamesVasseur: Agree with you there

matttsss:
@andrepeiry @LouisJamesVasseur I hear your points on the serialization for the parking enums. However, I do not use these functions, they are only used internally to Room and the function signature is quite limited. The only thing I could do is "if the index is out of bounds", return a arbitrary enum value if that doesn't bother you?
Also, I did not bother with documenting those `@TypeConverter` functions since it is quite straight forward what they do, but I could add that if necessary.

app/src/main/java/com/github/se/cyrcle/model/parking/offline/OfflineParkingRepositoryRoom.kt
matttsss: It's inherited from the interfaces :)

LouisJamesVasseur (commented):
VERY good work overall. The length of the PR definitely makes sense and I think you did well to put all of this into 1 PR. Some light implementation comments on my end but I think @andrepeiry said most of what I wanted to say already so I'll leave it at that. Mind the comments and I will approve!

app/src/main/java/com/github/se/cyrcle/model/parking/offline/OfflineParkingRepositoryRoom.kt
LouisJamesVasseur: Yeah okay, I see. No need then

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
LouisJamesVasseur: Good implementation then!

LouisJamesVasseur:
> @andrepeiry @LouisJamesVasseur I hear your points on the serialization for the parking enums. However, I do not use these functions, they are only used internally to Room and the function signature is quite limited. The only thing I could do is "if the index is out of bounds", return a arbitrary enum value if that doesn't bother you? Also, I did not bother with documenting those `@TypeConverter` functions since it is quite straight forward what they do, but I could add that if necessary.

The arbitrary enum solution works for me, I think what you're suggesting makes sense. Understood on the documentation, honestly it was more out of concern about the documentation aspect of M3 since you added a lot of code here. I'll leave it up to your opinion if you want to add the doc you can otherwise it's non-blocking on my end

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
LouisJamesVasseur: I think the flatMap as is is appropriate

andrepeiry: Well then, let this stay as is, it was merely a suggestion ^^

andrepeiry:
> However, I do not use these functions, they are only used internally to Room and the function signature is quite limited. The only thing I could do is "if the index is out of bounds", return a arbitrary enum value if that doesn't bother you?

It's bothering me, as such exception could crash the app (I know it won't technically happens, as data corruption should be a quite negligible risk). Indeed, `if (corruption) return default` seems like a good enough solution for me ^^
If they are only used internally, maybe mark them as `private` ?

> Also, I did not bother with documenting those @TypeConverter functions since it is quite straight forward what they do, but I could add that if necessary.

I know documentation is as unloved as unit tests (for most human at least 😉), but it is always appreciated, and if I remember correctly, you are using GitHub Copilot, which is a great tool to autocomplete minimal documentation

matttsss:
Thank you both,  I have addressed your comments if you want to check.
@andrepeiry Concerning marking the `@TypeConverters` as `private`, that will not be possible since `Room` needs to access them from outside.

LouisJamesVasseur (approved):
Thanks for addressing the changes I requested! Great work!

andrepeiry (approved):
Thank you for addressing our comments, this PR is good to go !

PR #338, opened 2024-12-02
Add options to switch between online/offline mode

sean-prz:
DRAFT


app/src/main/java/com/github/se/cyrcle/model/network/ConnectivityObserver.kt
```diff
@@ -0,0 +1,80 @@
...
+      userViewModel.isOnlineMode.collect { isOnlineMode ->
+        if (isOnlineMode) onOnlineMode() else onOfflineMode()
+      }
+    }
+  }
+
+  /**
+   * Switches the app to online mode by
+   * - signing in the user anonymously
+   * - switching the map and parking viewmodels to online mode (Soon)
```
AlexandreBeeDeeBop, line 53: The coaches recommend to add TODO. (im not making the rules here)

app/src/main/java/com/github/se/cyrcle/model/network/NetworkReceiver.kt
```diff
@@ -0,0 +1,48 @@
...
+          _isConnected.value = isConnected
+        }
+      }
+
+  fun register() {
+    val networkRequest =
+        NetworkRequest.Builder().addCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET).build()
+    connectivityManager.registerNetworkCallback(networkRequest, networkCallback)
+  }
+}
```
AlexandreBeeDeeBop, line 48: Shouldn't we also have the following `fun unregister() {
      connectivityManager.unregisterNetworkCallback(networkCallback)
  }  ` ? Im not sure but it could prevent unnecessary system resource uses and maybe crashs that the callbacks could cause.

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -721,7 +730,29 @@ fun SettingsDialog(
...
           }
-
+          HorizontalDivider(
+              thickness = 1.dp,
+              modifier = Modifier.fillMaxWidth(),
+              color = MaterialTheme.colorScheme.onSurface.copy(alpha = 0.12f))
+          // Switch to Online/Offline Mode
+          Row(verticalAlignment = Alignment.CenterVertically, modifier = Modifier.fillMaxWidth()) {
+            Box(modifier = Modifier.width(64.dp), contentAlignment = Alignment.CenterStart) {
+              Switch(
```
AlexandreBeeDeeBop, line 740: Can you please add a `contentDescription` for accessibility purposes ?

app/src/main/java/com/github/se/cyrcle/ui/theme/atoms/Button.kt
```diff
@@ -386,8 +386,7 @@ fun GoogleSignInButton(onClick: () -> Unit) {
       shape = RoundedCornerShape(50),
       border = BorderStroke(1.dp, Color.LightGray),
       modifier =
-          Modifier.padding(start = 16.dp, end = 16.dp)
```
AlexandreBeeDeeBop, line 389: Why did you remove this ?

app/src/main/res/values/strings.xml
```diff
@@ -273,8 +274,12 @@
     <string name ="zone_selection_button_cancel">Cancel</string>
     <string name ="zone_selection_download_error">Error downloading zone</string>
 
-
-
+    <!-- Connectivity Alert Dialog -->
+    <string name="connectivity_alert_offline_title">No Internet Connection</string>
+    <string name="connectivity_alert_offline_message">Do you want to use your Offline Maps</string>
```
AlexandreBeeDeeBop, line 279: If this message is supposed to be a question, then a question mark is missing. Same for the online one.

app/src/main/java/com/github/se/cyrcle/model/network/ConnectivityObserver.kt
```diff
@@ -0,0 +1,80 @@
...
+import android.content.Context
+import android.util.Log
+import com.github.se.cyrcle.model.map.MapViewModel
+import com.github.se.cyrcle.model.parking.ParkingViewModel
+import com.github.se.cyrcle.model.user.UserViewModel
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+
+class ConnectivityObserver(
```
LouisJamesVasseur, line 12: Generally, for classes like this, a bit of documentation explaining what it is is appreciated! Non-blocking though

app/src/main/java/com/github/se/cyrcle/ui/theme/atoms/Button.kt
LouisJamesVasseur: Yeah, good question, plus it's used in a lot of places so need to be sure on this

LouisJamesVasseur (approved):
I see nothing much to add, feature tests well and the code is clean and well constructed. A couple inline comments I've added for you to check but none are blocking. Well done!

app/src/main/java/com/github/se/cyrcle/model/network/NetworkReceiver.kt
sean-prz: I tried having an unregister() but it was useless since we never want to unregister the listener.
 It will automatically stop when we close the app

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
sean-prz: For an image, ok if the user wants to know what the image represents, 
For an icon, ok if the user wants to know what does this icon represents.
But for a switch, the user will just see what the switch does with the text.
The text next to it acts as a content description, so no noeed.

app/src/main/java/com/github/se/cyrcle/ui/theme/atoms/Button.kt
sean-prz: That's probably a mistake then, I will put it back

app/src/main/res/values/strings.xml
sean-prz: You're right

app/src/main/java/com/github/se/cyrcle/model/network/NetworkReceiver.kt
AlexandreBeeDeeBop: I think when you kill the app android does it automatically actually. But if you dont kill the app, but let's say just go on another app while ours is still in the background, maybe you should unregister()...

sean-prz: Where should I call it then ?

app/src/main/java/com/github/se/cyrcle/model/network/ConnectivityObserver.kt
sean-prz: Done

app/src/main/java/com/github/se/cyrcle/model/network/NetworkReceiver.kt
sean-prz: Drop the case 🔫

AlexandreBeeDeeBop: 2 bucks

app/src/main/java/com/github/se/cyrcle/model/network/ConnectivityObserver.kt
AlexandreBeeDeeBop: this is not a TODO... a todo is in flashy yellow (therefore why it's recommanded) and you can write one with : 
//TODO "your_text"
And my remark also regards the function below which also needs it.

app/src/main/java/com/github/se/cyrcle/model/network/NetworkReceiver.kt
sean-prz: Damn even cheaper than your mom, nice

AlexandreBeeDeeBop: fuck you ni-

AlexandreBeeDeeBop (approved):
All my comments have been adressed. Good job, you can merge.

app/src/main/java/com/github/se/cyrcle/model/network/NetworkReceiver.kt
AlexandreBeeDeeBop: Nevermind, as long as there's no bugs, let's not bother. let's just keep that in mind in case we see weird behaviours sometime in the future.

AlexandreBeeDeeBop:
bruh

PR #342, opened 2024-12-03
Feature Search Bar for List Screen

Pijukai:
## Content of the PR

This PR introduces the following elements : 

1. Search Bar of the List Screen
2. A new Suggestion Menu for the List Screen
3. New String Formatting for the suggestion
4. My Location as a suggestion for List Screen

## How ? 

1. The Search bar appears when clicking on the corresponding button. This allows the bar to take less space in a screen that is already heavily concentrated. The Search Bar is an `OutlinedTextField`. 

2. The Suggestion Menu is a `DropDownMenu` because it fits well in the screen. It deploys below the search Bar and covers other infos so that the user can be fully concentrated on selecting the good suggestion. 

3. The formatting use the following address field : `publicName`, `road`, `city` and `country` (in this order of priority). It takes up to 3 of those fields and concatenate them together to get the `SuggestionFormatName`. This way we have the most complete yet short name. The list of suggestion use the `SuggestionFormatName` to filter out the suggestions, remove the duplicate adn display the suggestion. 

4.  My Location is a suggestion that is hardcoded and displayed if the user gave its location permission. This sets the cyrcle center to be the user's location. 

## Design Choices

Below the list of Design choices: 

- The suggestion Menu is different because I want it to cover the other infos on the `listScreen` so the user doesn't feel overwhelmed compared to the one in `MapScreen` (this Screen doesn't have a lot of infos)

- The Max length of the suggestion format for the `ListScreen` is 50 characters compared to `MapScreen` which is 70. This is a balance between the size of the suggestion Menu and wanting the suggestion to be complete and not truncated. 

- I take only 6 suggestions for the Suggestion Menu in the List Screen. More suggestions would break the `DropDownMenu`. Hence , the user is expected to write a complete address he has in mind and not a simple letter. 

- The last selected suggestion stays in the Search Bar in any case so that the user know what location is selected anytime

- Finally, to respect Nominatim API norm on number of requests, the request is sent when the user click on validate button not before. 

## Little Issue

This issues is coming from the ordering inside the list Screen. The Backend doesn't support the fact that userPosition and CyrcleCenter can be different (I am modifying the cyrcleCenter based on the suggestion selected). The ordering can be broken but the list of parkings itself is good. This will need a New PR and a New issue will be created. 

## Code Coverage

image
image

The global coverage for the project is 82%


app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
```diff
@@ -74,34 +96,68 @@ import com.github.se.cyrcle.ui.theme.atoms.ScoreStars
...
+  }
+
+  // value that says wether the user clicked on my Location Suggestion or not
+  val myLocation = remember { mutableStateOf(true) }
+
+  // location permission from location manager
+  val locPermission = permissionHandler.getLocalisationPerm().value
+
+  LaunchedEffect(userPosition, myLocation, chosenLocation.value) {
+    if (locPermission && myLocation.value) parkingViewModel.setCircleCenter(userPosition)
```
LouisJamesVasseur, line 143: I like documentation on the ifs, for example the logic of this isn't immediately obvious

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
```diff
@@ -74,34 +96,68 @@ import com.github.se.cyrcle.ui.theme.atoms.ScoreStars
...
+    else Log.e("SpotListScreen", parkingViewModel.closestParkings.value.toString())
+    parkingViewModel.setCircleCenter(
+        Point.fromLngLat(
+            chosenLocation.value.longitude.toDouble(), chosenLocation.value.latitude.toDouble()))
```
LouisJamesVasseur, line 147: what is the double's original type? I don't understand why casting needs to be done here...

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
```diff
@@ -160,8 +216,15 @@ fun SpotListScreen(
       }
 }
 
+@SuppressLint("StateFlowValueCalledInComposition")
```
LouisJamesVasseur, line 219: I've been told suppressing this isn't always the right move. To quote @matttsss, "the annotation is here to help you". collectAsStates are your friend!

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
```diff
@@ -525,3 +616,134 @@ fun ActionCard(
...
+ * Composable function that displays a search bar for the list screen.
+ *
+ * @param addressViewModel The ViewModel responsible for managing addresses.
+ * @param isTextFieldVisible A state that controls the visibility of the text field.
+ * @param myLocation A state that indicates whether the user has selected "My Location".
+ * @param chosenLocation A state that holds the address chosen by the user.
+ * @param permissionHandler The handler for managing permissions.
+ * @param textFieldValue The current value of the text field.
+ */
+@SuppressLint("StateFlowValueCalledInComposition")
```
LouisJamesVasseur, line 630: Same comment as above

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
```diff
@@ -525,3 +616,134 @@ fun ActionCard(
...
+                      suggestion.suggestionFormatDisplayName(maxSuggestionDisplayNameLengthList)
+                  if (displayName in seenNames) {
+                    false
+                  } else {
+                    seenNames.add(displayName)
+                    true
+                  }
+                }
+
+            for (address in uniqueSuggestions.value.take(6)) {
```
LouisJamesVasseur, line 731: magic number here. Make it a const val?

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
```diff
@@ -525,3 +616,134 @@ fun ActionCard(
...
+                    Text(address.suggestionFormatDisplayName(maxSuggestionDisplayNameLengthList))
+                  },
+                  onClick = {
+                    chosenLocation.value = address
+                    showSuggestions.value = false
+                    myLocation.value = false
+                    isTextFieldVisible.value = false
+                    textFieldValue.value =
+                        address.suggestionFormatDisplayName(maxSuggestionDisplayNameLengthList)
+                    focusManager.clearFocus()
```
LouisJamesVasseur, line 743: Just so I make sure I understand your whole code, what does this do and why is it here?

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
Pijukai: The original type here is a String that is why I had to cast it. Deserializing the value as a String was easier for me but i can try to change that if you want.

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
Pijukai: I will look into it i think it is a superflux line that isn't necessary now.

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
Pijukai: I will look into it as well.

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
Pijukai: Will do !

Pijukai:
I putted the Draft so you could see the feature itself but for comments that treat documentation and minor details just wait a bit if possible. The PR is finished and so is the code. I still have a few constants and some documentation and cleaning to do. I was waiting for Maria's PR to pass to rebase since she modifies FilterHeader. However thanks Louis for pointing out those little mistakes I will take care of them as soon as I can.

Pijukai:
By the way I have a question for you. Do you think I should add region as i relevant field for string formatting ? Remember that we want the suggestion to be as precise yet as short as possible.

LouisJamesVasseur:
Given our use case, I'm not 100% sure it's worth the tradeoff of adding region...

Pijukai:
All your comments should be fixed now @LouisJamesVasseur !

Pijukai:
I will add region for the formatting of the map since we can have longer name but not for the list Screen. I think it is a good trade off.

Armetiis:
> Given our use case, I'm not 100% sure it's worth the tradeoff of adding region...

I have to disagree on this. For example, in the US, the region field corresponds to the state. I think in that case, for example, it would be very important to indicate the region

LouisJamesVasseur:
> > Given our use case, I'm not 100% sure it's worth the tradeoff of adding region...
> 
> I have to disagree on this. For example, in the US, the region field corresponds to the state. I think in that case, for example, it would be very important to indicate the region. But I can see how in most cases it can be irrelevant. I am a bit unsure of what's the best decision either

You make a solid point. Really wish we could pattern match on what country was used but I don't think that's possible

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
LouisJamesVasseur: Thanks for clarifying!

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
LouisJamesVasseur: Non-blocking for me, so look into it if you wish. Don't think it'll cause any issues

app/src/main/java/com/github/se/cyrcle/model/address/Address.kt
```diff
@@ -98,4 +101,21 @@ data class Address(
...
   }
+
+  /**
+   * Function that format the Address to display its suggestion name.
+   *
+   * @param maxLength the maximum length of the string
+   * @param mode the mode of the suggestion (MapScreen for the map, ListScreen for the list)
+   * @return the suggestion formatted display name
+   */
+  fun suggestionFormatDisplayName(maxLength: Int, mode: String): String {
```
LouisJamesVasseur, line 112: Couldn't the mode have been an enum? I mean string works just fine but it feels like a more elegant solution for this sort of thing

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
```diff
@@ -50,43 +55,86 @@ import androidx.compose.ui.res.stringResource
...
 import com.github.se.cyrcle.ui.theme.molecules.BottomNavigationBar
 import com.github.se.cyrcle.ui.theme.molecules.FilterPanel
+import com.mapbox.geojson.Point
 import com.mapbox.turf.TurfMeasurement
 import kotlin.math.roundToInt
 
 const val CARD_HEIGHT = 120
 const val MAX_SWIPE_DISTANCE = 150
+const val maxSuggestionDisplayNameLengthList = 70
+const val NumberOfSuggestionsForMenu = 6
```
LouisJamesVasseur, line 80: This is an upper bound, right? In my tests I sometimes saw 3 suggestions for example. If so, could the name of this const reflect this?

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
LouisJamesVasseur: Well addressed! Thanks!

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
Pijukai: Yess it is a upper bound it takes 6 suggestion if the lsit is bigger than 6 otherwise it takes everything if the size is <=6

Pijukai:
> > Given our use case, I'm not 100% sure it's worth the tradeoff of adding region...
> 
> I have to disagree on this. For example, in the US, the region field corresponds to the state. I think in that case, for example, it would be very important to indicate the region. But I can see how in most cases it can be irrelevant. I am a bit unsure of what's the best decision either

That is why I added the region for the MapScreen to be as complete as possible since there is no constraints on the size whereas for the list Screen i decided not to use it. Think it works for you ?

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
```diff
@@ -50,43 +55,86 @@ import androidx.compose.ui.res.stringResource
...
+
+  LaunchedEffect(userPosition, myLocation, chosenLocation.value) {
+
+    // if suggestion MyLocatio is chosen and user has location permission then set the circle center
+    // to user position
+    if (locPermission && myLocation.value) parkingViewModel.setCircleCenter(userPosition)
+
+    // else if the user has chosen a location or hasn't given his location permission then set the
+    // circle center to the chosen location (default position is EPFL)
+    else Log.e("SpotListScreen", parkingViewModel.closestParkings.value.toString())
```
Armetiis, line 120: This log suggests an error. I think this should be a debug log with `Log.d` unless I haven't understood

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -595,11 +600,27 @@ fun SuggestionMenu(
...
+                    suggestion.suggestionFormatDisplayName(
+                        maxSuggestionDisplayNameLengthMap, "MapScreen")
+                if (displayName in seenNames) {
+                  false
+                } else {
+                  seenNames.add(displayName)
+                  true
+                }
+              }
+          items(uniqueSuggestions.value) { suggestion ->
```
Armetiis, line 623: This is nice. I like that you did this to remove duplicates

app/src/main/java/com/github/se/cyrcle/ui/theme/molecules/Filter.kt
```diff
@@ -263,3 +323,142 @@ fun FilterSection(
...
+    if (listOfSuggestions.value.size + 1 > 0) {
+      DropdownMenu(
+          expanded = showSuggestions.value,
+          onDismissRequest = { showSuggestions.value = false },
+          modifier =
+              Modifier.width(with(LocalDensity.current) { textFieldSize.value.width.toDp() })
+                  .testTag("SuggestionsMenuListScreen")) {
+            if (permissionHandler.getLocalisationPerm().collectAsState().value) {
+              DropdownMenuItem(
+                  text = { Text("My Location") },
```
Armetiis, line 417: Here you forgot to use a string resource, which you could reuse on line 422

app/src/main/java/com/github/se/cyrcle/ui/theme/molecules/Filter.kt
```diff
@@ -263,3 +323,142 @@ fun FilterSection(
...
+                  onClick = {
+                    chosenLocation.value = address
+                    showSuggestions.value = false
+                    myLocation.value = false
+                    isTextFieldVisible.value = false
+                    textFieldValue.value =
+                        address.suggestionFormatDisplayName(
+                            maxSuggestionDisplayNameLengthList, "ListScreen")
+                    focusManager.clearFocus()
+                  })
```
Armetiis, line 459: Here i think for the readability of the suggestions it is very important to add `contentPadding = PaddingValues(8.dp)`

app/src/main/java/com/github/se/cyrcle/model/address/Address.kt
Pijukai: I added the enum and indeed it makes the code cleaner!

Armetiis (changes requested):
Hi. I tried to dive really deep into this PR to try to see all the improvements that can be made and here is what I'm left with:

First of all, good job. You did very good, I am very satisfied with how this fits on the List Screen. You managed to not make it too packed and agressive to the sight. I love it.

While testing I noticed 2 errors and an improvement that can be made that aren't proper of this PR but are closely related to the output of the search bar.

1. This is my mistake, from when I conceived the Address dataclass. The serialized name annotations of some attributes are incorrect. For example:
```kotlin
@SerializedName("city", alternate = ["municipality", "village, town"]) val city: String = ""
```
You can see that there is a problem with the `"village, town"` instead of ` "village", "town"`. This caused the village and town to not be recognized and typing for example "Chavannes Pres Renens" would only output "Schweiz/Suisse/Svizzera/Svizra" (which is really annoying and long btw but that will come in point n.3). Here is a patch where i fixed those (as this is my bad)
[0001-fix-fix-serialized-name-annotations.patch](https://github.com/user-attachments/files/18005984/0001-fix-fix-serialized-name-annotations.patch). Please double check that I didn't miss one, even if I already triple checked

2. This was introduced when the search bar was added to the map. The error is in the `addressFromBody` function. I copied the problematic code snippet with a FIXME on the line that causes the problem
```kotlin
private fun addressFromBody(body: String): List<Address> {
    val jsonArray = JSONArray(body)
    if (jsonArray.length() == 0) {
      return listOf(Address()) // FIXME: HERE WE SHOULD RETURN EMPTY LIST
    }
    //...
  }
```
Notice that this means that when there are no results, we display an empty Address and clicking on it takes us to the point 0.0, 0.0. (In the map it's even worse because we don't even have to confirm, it automatically takes us there).

3. Finally, this is an improvement I propose. As we said before, having "Schweiz/Suisse/Svizzera/Svizra" is REALLY annoying. What I suggest is adding the following to the query parameters in `search` in `AddressRepositoryNominatim` (commented line with TODO)

```
val url =
        HttpUrl.Builder()
            .scheme("https")
            .host("nominatim.openstreetmap.org")
            .addPathSegment("search")
            .addQueryParameter("q", query)
            .addQueryParameter("format", "json")
            .addQueryParameter("addressdetails", "1")
            .addQueryParameter("accept-language", "en") // TODO: add this line
            .build()
```
 Then it will only suggest "Switzerland" rather than all languages. I think this is very useful to make the suggestions more lightweight.
 
 Coming back to what you said in "Little issue". Can't you create a public stateFlow circleCenter that you can use in the list screen to compute the distance. I think the sorting is correct, the only problem is the line 155 of the ListScreen: `val distance = TurfMeasurement.distance(userPosition, parking.location.center)` where instead of userPosition you should use circle center. It's not a backend problem, just a UI one. Please correct me if you think I am wrong

Finally, I also left you some comments in the code, please make sure to consider them.

app/src/main/java/com/github/se/cyrcle/model/address/Address.kt
```diff
@@ -98,4 +101,21 @@ data class Address(
...
+    val fieldPriorities: List<String>
+    if (mode == "ListScreen") {
+      fieldPriorities = listOf(publicName, road, city, country)
+    } else {
```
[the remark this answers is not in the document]
Armetiis: Well I guess this was already discussed and fixed while I was writing my review. Good job

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
```diff
@@ -50,43 +55,86 @@ import androidx.compose.ui.res.stringResource
...
+        Address(
+            latitude = MapConfig.defaultCameraState().center.latitude().toString(),
+            longitude = MapConfig.defaultCameraState().center.longitude().toString()))
+  }
+
+  // value that says wether the user clicked on my Location Suggestion or not
+  val myLocation = remember { mutableStateOf(true) }
+
+  // location permission from location manager
+  val locPermission = permissionHandler.getLocalisationPerm().value
```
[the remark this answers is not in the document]
Pijukai: You are right I changed this but due to the rebasing it must have disappeared. Will do !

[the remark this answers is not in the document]
Pijukai: done !

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
Pijukai: This is the case eveywhere in the code but i will still do it if it matter for you

app/src/main/java/com/github/se/cyrcle/ui/theme/molecules/Filter.kt
Pijukai: done !

app/src/main/java/com/github/se/cyrcle/ui/theme/molecules/Filter.kt
Pijukai: fixed !

app/src/main/java/com/github/se/cyrcle/ui/theme/molecules/Filter.kt
```diff
@@ -263,3 +323,142 @@ fun FilterSection(
...
+                  } else {
+                    seenNames.add(displayName)
+                    true
+                  }
+                }
+
+            for (address in uniqueSuggestions.value.take(NumberOfSuggestionsForMenu)) {
+              DropdownMenuItem(
+                  text = {
+                    Text(
```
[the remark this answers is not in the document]
Pijukai: Fixed as well !

app/src/main/java/com/github/se/cyrcle/ui/theme/molecules/Filter.kt
```diff
@@ -263,3 +323,142 @@ fun FilterSection(
...
+  val slideOffset by
+      animateDpAsState(targetValue = if (isTextFieldVisible.value) 0.dp else (-200).dp)
+
+  Box {
+    OutlinedTextField(
+        value = textFieldValue.value,
+        onValueChange = { textFieldValue.value = it },
+        placeholder = { Text(text = stringResource(R.string.search_bar_placeholder)) },
+        modifier =
+            Modifier.padding(start = 8.dp)
```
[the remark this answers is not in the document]
Pijukai: I think the text field is okay as it is i don't see the need for the padding here. Is it a blocking issue for you ?

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -120,6 +120,7 @@ const val LAYER_ID = "0128"
 const val LAYER_ID_RECT = "0129"
 const val ADVANCED_MODE_ZOOM_THRESHOLD = 15.5
 const val CLUSTER_COLORS = "#1A4988"
+const val maxSuggestionDisplayNameLengthMap = 100
```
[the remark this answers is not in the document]
Pijukai: fixed !

Pijukai:
I fixed everything as you suggested and it improves the style and the suggestions so thanks for the feedback ! However for the improvement that you mention, it don't think this solution is a good idea. the solution consists of bringing logic to the UI to fix a structural problem in the view model. So I think this fix is the subject of a PR in itself where the modification should be done with how we manage the user_position and cyrcleCenter. An issue as been or will be opened for the matter. Moreover I think this PR covers a lot (maybe too much) code/topics compared to what a traditional PR should cover.

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
Armetiis: Really? I never noticed 🤭

app/src/main/java/com/github/se/cyrcle/ui/theme/molecules/Filter.kt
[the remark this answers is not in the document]
Pijukai: Now it should be fixed. I didn't understand your comment but changing emulator made me realize why you brought that up.

Armetiis (approved):
Perfect, thank you so much for addressing the changes. Please for the problem we talked about, make sure to create an issue describing what the problem is and what's your proposed solution, so we can discuss it and assign it for next sprint

It's very very good, I'm very happy with how much the changes improved the feature, and how the final result looks and works.

Pijukai:
I opened an issue (#347) and described the problem

Issue #276, opened 2024-11-22
BugFix : Re-enable rotation on the map 


Pijukai:
Re-enable the rotation on the Map. Need to watch for excess of requests to firebase.

PR #345, opened 2024-12-04
Re-enable rotation on the Map and display parkings in Location Picker screen

### LINKED ISSUE #320: Re introduce the parkings in the Add Parking Screen map
We can't see the other parkings when adding one because of the surge of requests to the firebase since the map can be rotated.

Pijukai:
## Content of the PR 

This PR does the following : 

1. Re-enable the rotation on the map 
2. Display the parkings on the `LocationPicker` screen
3. Add debugging tags to track real time number of request to database since the app launched

## How ?

1. Removed one line in `MapScreen` that I added as a safety measure when backend couldn't handle rotation of the map
2. I added a `LaunchedEffect` with attribute `listOfParkings`. This way when the attribute is updated the `LaunchedEffect` draw the makers using the `markerAnnotationManager`
3. Added a private var `nbrOrRequest` that is increased by one each time a `get()` function of the `ParkingRepositoryFirestore` is called. The real time value is also displayed through a Log for each `get()` method. 

## Additionnal Infos

I removed the file `ZoomControls.kt` which is not useful anymore since my last PR on the search bar for the Map. It will not be useful in the future and makes the code coverage drop. 

## Code Coverages 
image
The modified files are : 
- Add parking location overlay at 82%
- Add parking location 91% 
- Map which is 77% (unfortunately due to the same problem but the gap is getting smaller)
## Closing issues

Closes #320 #276


sean-prz:
The features seems to work great congrats.
I am a bit annoyed that the cardinals points icon is hidden behind the settings button, making it unaccessible, I can not reset my rotation to have a map aligned with the north.
Could you take a look to see if we can modify the icon position ?

Pijukai:
> The features seems to work great congrats. I am a bit annoyed that the cardinals points icon is hidden behind the settings button, making it unaccessible, I can not reset my rotation to have a map aligned with the north. Could you take a look to see if we can modify the icon position ?

Sure i will take a look at it .

Pijukai:
I tried but I am not able to find the cardinal points icon nor able to fix this without spending time on it. I already spent a lot of time this week so i will stick to the subject of this PR which was to re enable the rotation. I will make sure create an issue and bring it up during the planification meeting for the polishing.

Pijukai:
here is the issue #349

app/src/main/java/com/github/se/cyrcle/model/parking/online/ParkingRepositoryFirestore.kt
andrepeiry, line 1: I don't think you meant to log errors in here (`Log.e`). Maybe you meant information (`Log.i`) or verbose (`Log.v`)

Pijukai: fixed thanks you for the comment !

andrepeiry: Completely forgot there was `Log.d` while looking at the documentation ^^'

andrepeiry (approved):
# Nice PR and code !
Code was pleasant to read, and the description true to its content.
As for the hidden compass, I think this is better to leave for the polish sprint or simply for next sprint.

## Task separation
One could argue that this should have been two (given the difference in tasks), but as I think the size of your PR, but I won't complain a lot as they are slightly linked on the map subject, I won't make it a big deal 😉

## When you will work at a FAANG company
If I wanted to be as picky as one could get, I would tell you to add some tests to assert there are no more Zoom controls on the map, as the feature was deprecated, but it isn't useful for our scale.

andrepeiry:
> The link between the tasks was that the display of parking was deactivated due to the rotation being not functionnal. As for the zoom controls i think deleting the file itself make me sure that it is not displayed. But I can still add tests if you want.

I see the link no, thank you ^^

Do not worry about the zoom controls, it was merely an extension as to ensure "The app is what we expect it to be", but it was so overkill that I left is as a comment

app/src/main/java/com/github/se/cyrcle/ui/addParking/location/LocationPicker.kt
```diff
@@ -35,6 +45,21 @@ fun LocationPicker(
...
   val locationPickerState by mapViewModel.locationPickerState.collectAsState()
+
+  // Mutable state to store the PolygonAnnotationManager for rectangles
+  var rectangleAnnotationManager by remember { mutableStateOf<PolygonAnnotationManager?>(null) }
+
+  // Mutable state to store the PointAnnotationManager for parking labels
+  var pLabelAnnotationManager by remember { mutableStateOf<PointAnnotationManager?>(null) }
+
+  // Collect the list of parkings from the ParkingViewModel as a state
+  val listOfParkings by parkingViewModel.filteredRectParkings.collectAsState(emptyList())
```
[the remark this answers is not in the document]
Pijukai: You are right thank you for taking care of it !

PR #346, opened 2024-12-04
Increase User integration into Reports

LouisJamesVasseur:
### What is this PR?

This PR introduces a few functionalities to make Users more central to the Reporting facet of Cyrcle.

### Why?

Reports are, of course, strongly about Users who are trying to make their voices heard about the content they don't want to see in the app. It's therefore important that they can do so in a system that's coherent with the User architecture.

### How?

3 things were mainly added
- Users can now only submit 1 Report per Object (Parking, Review and soon Image) to add weight to the Reports that Users submit
- A ReportedObject is now associated to an owner, which was made possible by associating a Parking to an owner which wasn't done before
- The Admin user is now defined, which limits access to the Admin Screen

### Tests

A couple new ViewModel tests were added


app/src/main/java/com/github/se/cyrcle/model/review/ReviewViewModel.kt
```diff
@@ -268,6 +283,32 @@ class ReviewViewModel(
     Log.d("ReviewViewModel", "Review and metrics updated successfully: ${selectedReview.nbReports}")
   }
 
+  fun addReportingUser(user: User) {
```
andrepeiry, line 286: Missing documentation.

app/src/main/java/com/github/se/cyrcle/model/review/ReviewViewModel.kt
```diff
@@ -268,6 +283,32 @@ class ReviewViewModel(
...
+    _selectedReview.update { currentReview ->
+      currentReview?.let {
+        val updatedReview = it.copy(reportingUsers = it.reportingUsers + user.public.userId)
+
+        reviewRepository.updateReview(
+            review = updatedReview,
+            onSuccess = {
+              Log.d("ReviewViewModel", "User added to reportingUsers in Firestore")
+              _selectedReview.update { state ->
```
andrepeiry, line 295: Maybe I'm mistaken here, but I don't think that `state` is a good name.
The below `if (...) updatedReview else state` feels really weird to read, so I would go for ~`state`~ to `currentReview`.
Also, should it be `if (...) { update{} }` instead of ~`update { if (...) {} else {} }`~, because the current implementation might trigger some listener chain for nothing, hence why I would think to put the `if` outside of this `update` lambda.

andrepeiry (commented):
Nice PR in general, but a small blocking part in the comments I left you inline

andrepeiry (approved):
Thank you for addressing my comments. Approved !

andrepeiry (approved):

sean-prz (changes requested):
I don't think it's the expected behaviour, when I add a report, the list of user who reported the object is not updated/(created) and so currently I can still add multiple reports! Take a look and come back to me

LouisJamesVasseur:
Will do! I am actively looking into it thanks (I think I messed something up when addressing André's request)

app/src/main/java/com/github/se/cyrcle/model/review/ReviewViewModel.kt
LouisJamesVasseur: I've spent a few hours trying to get an alternative to work and it unfortunately hasn't as of yet. I've reverted this branch to the state it was at before I addressed your change. Is this really blocking for you? I think, given the stateFlows that I'm using that the permanent update is necessary, at least that's what my tests strongly suggest

LouisJamesVasseur:
Hi reviewers, thanks for bearing with me while I got that sorted out.

Here's what the issue was: for some reason I'm still not 100% sure of, the ParkingViewModel and its backend don't update in the same way that the other ViewModels do, meaning that my solution using a state and attaching reportingUsers to Parkings wasn't working (PS: the way I found this out was by reworking the addReportingUser function at André's suggestion so thanks!). My fix for this was to make it the other way around and attach ReportedParkings to the Users directly. Not only does this work, but it also makes the code a lot less bloated and much cleaner in my opinion (I hope you will agree). I am now confident in re-opening this PR!

sean-prz (changes requested):
There is a fatal crash when trying to open the details screen for a parking with no owner

LouisJamesVasseur:
> There is a fatal crash when trying to open the details screen for a parking with no owner
Screenshot 2024-12-05 at 10 52 20

Seems an odd bug is happening...

sean-prz:
Crash is fixed, nice ! Sometimes it shows nothing instead of "Unknown user" for old parkings, any idea why ?
image

sean-prz (changes requested):
image
Still not sure this is the wanted behaviour

LouisJamesVasseur:
> image Still not sure this is the wanted behaviour

It actually was. The whole point was to show, on the UI, that Parkings had users now. I hear you that this is not sufficient and you want to username to be displayed and am working towards a fix

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
andrepeiry, line 1: Missing documentation on several functions in this file.

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
```diff
@@ -88,6 +88,11 @@ fun ParkingDetailsScreen(
   val showDialog = remember { mutableStateOf(false) }
   val imagesUrls by parkingViewModel.selectedParkingImagesUrls.collectAsState()
   val showDialogImage = remember { mutableStateOf<String?>(null) }
+  if (selectedParking.owner != "Unknown Owner" && selectedParking.owner != null) {
```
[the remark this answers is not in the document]
LouisJamesVasseur: It's possible. Since there are only two distinct cases here I don't think creating the function itself would be too worth it for now, I'll keep it in mind if this comes back up

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
LouisJamesVasseur: I've added some where needed in a recent commit

sean-prz (approved):
Finally approved ! We can't add multiple reports anymore and there is no fatal crashes!

As you can see the username doesn't display on first launch of a screen, but this is an acceptable problem for now : 

[Screen_recording_20241205_161759.webm](https://github.com/user-attachments/assets/432fb442-dbcb-4c1f-a786-c25dda6b75fa)

andrepeiry (approved):

LouisJamesVasseur:
> Finally approved ! We can't add multiple reports anymore and there is no fatal crashes! 🥳
> 
> As you can see the username doesn't display on first launch of a screen, but this is an acceptable problem for now :
> 
>  Screen_recording_20241205_161759.webm

Oddly enough, I can't reproduce the behavior. I'll have a look later but I agree that the feature works acceptably for what I wanted this PR

PR #351, opened 2024-12-04
Polish the MapBoxMap

sean-prz:
- fix : #349 
## What
- Removes the scaleBar/Ruler that was behind the search bar but sometimes overflowing
- Moves the compass icon under setting button to reset the rotation
- Display ALL parkings instead of filtered one on the location Picker
# Before 
image
# After
image


Pijukai (approved):
The compass was in the parameter of the map well done Sean ! All the time spent to correct the null pointer exception xD. Anyway Approved good job!

Armetiis (approved):
This is nice. Thank you

PR #352, opened 2024-12-04
Bugfix 343: Replace XML-based preview cards with Composable overlay

### LINKED ISSUE #343: Map screen preview cards not displaying upon marker click
In some situations, clicking a marker centers the camera around it but does not display the card. See for example the video below

https://github.com/user-attachments/assets/408cfcc6-4255-4e2d-8eba-e9d0ed049987

Now the cause of the bug is not identified yet. How to replicate it is not very clear either, but it is usually triggered when:
- Like in the video above, we change the zoom level and try to click on markers
- Apply filters on mapscreen also triggers the bug

Note that it is not related to the zoom levels nor the filters themselves, as going for example to list screen and coming back to map screen fixes the bug

Armetiis:
This PR addresses an intermittent issue where preview cards, previously implemented as XML resources, were not consistently displayed when the markers were clicked. This solution replaces the XML-based approach with a Composable overlay, improving reliability and maintainability.

### Changes

1. **Removed ViewAnnotationManager in the MapViewModel:**
   - Eliminated `ViewAnnotationManager` and related code.
   - Removed calls to `mapViewModel.updateViewAnnotationManager()` and `mapViewModel.removePreviewCard()`.

2. **Introduced Composable PreviewCard:**
   - Created a new `PreviewCard` Composable function to display parking information.
   - The card is positioned using a custom layout modifier
   - Introduced `showPreviewCard` state to control the visibility of the preview card.
   - Updated click listeners to set `showPreviewCard = true` when a parking is selected.

3. **Refactored Click Handlers:**
   - Created a common `handleAnnotationClick` function for both point and polygon annotations.
   - This function updates the selected parking, shows the preview card, and centers the map.
   
### Benefits

- **Improved Reliability:** Eliminates issues related to XML-based view annotations not displaying correctly.
- **Enhanced Information:** Users now see more details about each parking spot in the preview.
- **Modification potential:** Having a composable instead of a string resource makes it much easier to customize the card

### New Preview Card's UI
image

### Testing and coverage
As you all know, it is not possible to interact with the map in the tests to select a parking. This makes it impossible to test the interaction with the markers and ensure the correct dispaly of the card. However, given that the card is now a composable, it was possible to test it in isolation by setting it as content of the compose test rule.
image

-- -
- Closes #343

### DIFF
--- a/app/src/androidTest/java/com/github/se/cyrcle/ui/map/MapScreenTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/ui/map/MapScreenTest.kt
@@ -23,6 +23,7 @@ import com.github.se.cyrcle.di.mocks.MockUserRepository
 import com.github.se.cyrcle.model.address.AddressViewModel
 import com.github.se.cyrcle.model.map.MapViewModel
 import com.github.se.cyrcle.model.parking.ParkingViewModel
+import com.github.se.cyrcle.model.parking.TestInstancesParking
 import com.github.se.cyrcle.model.report.ReportedObjectRepository
 import com.github.se.cyrcle.model.user.TestInstancesUser
 import com.github.se.cyrcle.model.user.UserViewModel
@@ -284,5 +285,26 @@ class MapScreenTest {
         .onNodeWithTag("suggestionCardMock City 2")
         .assertIsDisplayed()
         .assertHasClickAction()
+  }
+
+  @Test
+  fun testPreviewCard() {
+    composeTestRule.setContent {
+      PreviewCard(navigationActions = mockNavigation, parkingViewModel = parkingViewModel)
+      parkingViewModel.selectParking(TestInstancesParking.parking1)
+    }
+    composeTestRule.waitForIdle()
+
+    composeTestRule.onNodeWithTag("PreviewCard").assertIsDisplayed()
+    composeTestRule.onNodeWithTag("PreviewCardTitle").assertIsDisplayed()
+    composeTestRule.onNodeWithTag("PreviewCardRackType").assertIsDisplayed()
+    composeTestRule.onNodeWithTag("PreviewCardProtection").assertIsDisplayed()
+    composeTestRule.onNodeWithTag("PreviewCardCapacity").assertIsDisplayed()
+    composeTestRule
+        .onNodeWithTag("PreviewCardDetailsButton")
+        .assertIsDisplayed()
+        .assertHasClickAction()
+        .performClick()
+    verify(mockNavigation).navigateTo(Screen.PARKING_DETAILS)
   }
 }
--- a/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
@@ -397,6 +397,12 @@ fun MapScreen(
                     mapViewModel.updateMapMode(userMapMode.value)
                   }
 
+                  // If the zoom level changes, hide the preview card. This is to avoid the card
+                  // being displayed when the user taps to zoom in/out.
+                  if (zoomState.value != mapView.mapboxMap.cameraState.zoom) {
+                    showPreviewCard = false
+                  }
+
                   // Store the zoom level
                   // This must stay at the end of the listener.
                   zoomState.value = mapView.mapboxMap.cameraState.zoom
--- a/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
@@ -770,7 +770,8 @@ fun PreviewCard(navigationActions: NavigationActions, parkingViewModel: ParkingV
                     val yOffset = -placeable.height / 2 - 50.dp.roundToPx()
                     placeable.placeRelative(0, yOffset)
                   }
-                },
+                }
+                .testTag("PreviewCard"),
         colors =
             cardColors(
                 containerColor = MaterialTheme.colorScheme.surface,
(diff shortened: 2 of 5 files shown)


Pijukai:
First i want to say that the preview card looks great now so good job. Second we finally passed the 80% in map so congrats as well.

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -329,89 +325,46 @@ fun MapScreen(
...
-                      mapViewModel.removePreviewCard()
-                      // get the data from the PointAnnotation and deserialize it
-                      val parkingData = pointAnnotation.getData()?.asJsonObject
-                      val parkingDeserialized = gson.fromJson(parkingData, Parking::class.java)
-                      // move the camera to the selected parking
-                      mapViewportState.setCameraOptions {
-                        center(parkingDeserialized.location.center)
+                // ======================= ANNOTATION CLICK LISTENER =======================
+                // Define a common click handler function
+                fun handleAnnotationClick(
```
Pijukai, line 330: since it is a function maybe you can add the description for the parameters ?

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -329,89 +325,46 @@ fun MapScreen(
...
+                    annotation: Any,
```
Pijukai, line 331: Here instead of using `Any` you could have used `Annotation` since it is the superclass of both PointAnnotation and PolygonAnnotation. This way we are more in control of what goes inside the function.

app/src/androidTest/java/com/github/se/cyrcle/ui/map/MapScreenTest.kt
```diff
@@ -285,4 +286,25 @@ class MapScreenTest {
         .assertIsDisplayed()
         .assertHasClickAction()
   }
+
+  @Test
+  fun testPreviewCard() {
```
Pijukai, line 291: You did the best that you could to test it so well done :+1:

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -163,6 +164,10 @@ fun MapScreen(
   // Mutable state to store the PointAnnotationManager for parking labels
   var pLabelAnnotationManager by remember { mutableStateOf<PointAnnotationManager?>(null) }
 
+  var showPreviewCard by remember { mutableStateOf(false) }
```
Pijukai, line 167: It is a non blocking issue but since all other variables have documentation maybe you can add one here ?

app/src/androidTest/java/com/github/se/cyrcle/ui/map/MapScreenTest.kt
Pijukai: You could improve the test by asserting what is displayed on the card since it is linked to parking1 if I am not wrong.

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Armetiis: I thought about that, but somehow it doesn't work. I get the following error: Incompatible types: PointAnnotation and Annotation. I don't really know how to do it

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Armetiis: Sure, i'll do it for consistency. But I believe that sometimes, trivial comments should be avoided. The name of the variable is self explanatory.

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Armetiis: Sure, will add documentation on this

app/src/androidTest/java/com/github/se/cyrcle/ui/map/MapScreenTest.kt
Armetiis: About this. The content of the preview card is clearly not the definitive one. I just did this to have an idea of what the Preview Card will look like. The point of this PR is not to design a Preview Card, but rather to ensure their reliability. This is something that can be addressed when we work on its content during the polish sprint.

sean-prz (commented):
Thanks for this ! Another PR down the drain fixing problems with the map !
This will make adding the icons on the card next sprint so much easier. 

The code is clean, I will not waste our time on futilities. 
Instead take the time to fix and relevant problems : 

When we zoom with double click the card is not removed, and so we can move out and the card floats around without being on top of the parking : 



If we filter out the parking the card stays.

You can take advantage of the zoom listener to fix the first problem and the launchedEffect on the filteredParkings for the second !
Hf

Armetiis:
> When we zoom with double click the card is not removed, and so we can move out and the card floats around without being on top of the parking 

Fixed! Now if the zoomLevel changes, the card gets disabled


> If we filter out the parking the card stays.

I changed this. Now the preview card is disabled if the parking gets filtered out from the selected parking, but it doesn't if the list gets updated but the parking is still in it :)

Thanks for the suggestions. Little details like this make it perfect!

sean-prz (approved):
Thank you, Perfect !

app/src/androidTest/java/com/github/se/cyrcle/ui/map/MapScreenTest.kt
Pijukai: It was just a suggestion i agree with you and i will make sure to bring it up during the polishing.

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
Pijukai: The code is good as it is and the switch will handle this perfectly so it doesn't matter for this PR. Since you looked into i will not block the PR for it.

Pijukai (approved):
Thank you for fixing this bug ! The overall quality of the PR is really good and apart from the bug even the cc goes above 80% for the map. Great Job !

PR #350, opened 2024-12-04
Show in Map button

andrepeiry:
# Content of this PR

## How
- Modify the `onClick` lambda in the button "Show in Map" in the `ParkingDetailsScreen`.
- Add a function in MapViewModel to zoom on a location. Allow to move the camera without getting the `_cameraState`.
- ParkingDetailsScreen now require the MapViewModel to navigate to a particular position.

## Small note
The documentation of the `MapViewModel` , while existing, was a bit hard to read. Maybe some clarification could be done in the polish sprint.

### Changes in code coverage
- `ui/parkingDetails` 84% to 84%
- `model/map` 82% to 82%
- *general* 82% to 82%

## Close
Close #348

PR #359, opened 2024-12-05
Feature go to map button Bis

andrepeiry:
This PR is merely a copy of  #350, with the only difference being inside the AndroidTest `ParkingDetailsScreenTest#showInMapButtonGoesToMapScreen` because the CI did not pass previously.


matttsss (commented):
An overall functioning PR and it's nice to have this button working. However, I'm not a fan of the idea of just commenting a problematic test. Would it not be possible to rescale the UI for the "Small Phone" emulator and keep the tests as they were?

AlexandreBeeDeeBop (approved):
If you really can't find a way to make this test pass on the CI, then as long as the code coverage is good enough ill let it slide

matttsss (approved):
All good with me! Thanks for adapting the UI!

PR #372, opened 2024-12-06
Improve serializers when dealing with legacy

matttsss:
## What is this PR?

This PR aims to reduce bugs caused by legacy parkings that lack certain attributes in Firestore (the gson deserializer sets it to `null` and that causes `NullPointerExceptions`).

### How?
By adding checks to the newer attributes when deserializing parkings.

### Code coverage:
NO NEED MOUHAHAAHAHAHAH


sean-prz (approved):
Thank you for your reactivity, this fixed the fatal crash I was facing, and doesn't seem to produce any new ones

LouisJamesVasseur (approved):
Awesome, thanks for the reactive fixes!

PR #373, opened 2024-12-07
Feature offline parkings

### LINKED ISSUE #318: Add state for offline mode in the ParkingViewModel
Parkings should be fetched from the correct source.

matttsss:
## What is this PR?

This PR aims to finalize the `ParkingViewModel` for offline mode. It adds a "switch" that selects the repository to get parkings from depending on the State of the connection.

### What does it mean?

This implies that parkings can now be fetched from the downloaded zones when the phone is offline!

Closes #318

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/model/network/ConnectivityObserver.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/network/ConnectivityObserver.kt
@@ -71,7 +71,7 @@ class ConnectivityObserver(
           userViewModel.setIsOnlineMode(false)
         })
     // mapviewmodel.switchToOnlineMode()
-    // parkingViewModel.switchToOnlineMode()
+    parkingViewModel.switchToOnlineMode()
   }
 
   /**
@@ -83,6 +83,6 @@ class ConnectivityObserver(
     Log.d("ConnectivityObserver", "Switching to offline mode")
     userViewModel.setCurrentUser(null)
     // mapviewmodel.switchToOfflineMode()
-    // parkingViewModel.switchToOfflineMode()
+    parkingViewModel.switchToOfflineMode()
   }
 }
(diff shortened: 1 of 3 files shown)


sean-prz:
Hey, How can I make sure I am not loading the cached data but really the data from our local db, stupid question probably but I didn't manage to disable the cache. 

How do we test the feature ?

matttsss:
> How do we test the feature ?
You can download a zone and then turn to offline mode, it will flush the cache of our LinkedHashMap and swap repositories, you can then turn back to online mode as you wish.
There is a little latency when loading parkings after a swap of about a second or two though.

sean-prz (approved):
Thanks for the clarification.

I tested downloading and deleting some zones (Annecy, Lausanne, Genève) and switching between modes and everything seems to work perfectly. 

Good job ! you did it !

Armetiis:
Hi. I have to be honest I am getting some unexpected behaviours.

First, make sure to have no downloaded zones on your device (at least not a tile containing lausanne)
Login Offline
Switch back to online, and wait for parkings to load
Switch back to offline. Parkings don't get flushed until you change screen (go to list and back to map)
If I switch back online again at this point, parkings don't load until I switch screen

Another thing I encountered: I started downloading Lausanne, switched to offline. I confirmed that only parkings in lausanne remained. However, when switching back to online mode it took waay more than a couple seconds ~~(something like 1 minute)~~ to display those in Neuchatel. Then somehow, after restarting the app (online, without ever touching the online/offline switch), ~~they took like a whole minute to load.~~ Actually not quite. I have to switch screen + come back to map then wait something around 2 seconds

matttsss:
@Armetiis Good catch! I didn't flush all parking caches in the VM correctly. Concerning UI updating of the `MapScreen`, in the current state, changing connection mode does not trigger an UI update on the MapScreen, you would have to do so by panning around to force update. With regards to load times, not much can be done about that, we are dependent on the framework unless we do overkill performance optimizations.

I have made additional issues to track that:
* open: #386 
* open: #387

sean-prz (approved):

Armetiis (approved):
I have nothing to add. I confirmed that the bug i was experiencing was not related to your PR and it was just related to the zoom treshold for fetch, and that this is related to the issue you created for optimization. 
Good job!

PR #378, opened 2024-12-08
Feature offline mapbox

sean-prz:
## What

This PR ensures the following : 
- The Mapbox tiles are now downloaded when the user adds a new zone
- The Mapbox tiles are deleted when the user deletes a zone
- When Mapbox can't fetch the tiles from the server, it will automatically uses the tiles in the storage. 

## How

**Disclaimer** I mostly followed and copied the documentation from MapBox on how to do this, so if you want to understand whats happening read : https://docs.mapbox.com/android/maps/guides/offline/
#### Mapconfig.downloadZone() & Mapconfig.deleteZone()

This functions are  in MapConfig, rather than MapviewModel,
I felt like it didn't belong in a Viewmodel, since it doesn't have any states or any interactions with the UI.

#### Managing downloaded zones 
These two functions are called when the user adds/refreshes/deletes a new zone, from the zone manager/selection screens.

#### Switching between online/offline tiles
This is done automatically by Mapbox, 

## Why
This is the last PR to make our offline mode functional on the side of the map data.

## Note for reviewers

You can test the feature by downloading a zone and then reloading the app offline and check that the area is visible, downloaded in high quality.

To make the distinction between the cached data and persistent storage you can add this code snippet that deletes the cache (!= the downloaded tiles from the zones)

In the global settings section of the MapScreen : 

```
TileStore.create().clearAmbientCache { expected ->  
  if (expected.isValue) {  
    Log.d("MapScreen", "Cache cleared successfully")  
  } else {  
    Log.e("MapScreen", "Error clearing cache: ${expected.error}")  
  }  
}
```

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/map/MapConfig.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/map/MapConfig.kt
@@ -42,6 +42,7 @@ object MapConfig {
     return MapStyle("mapbox://styles/seanprz/cm27wh9ff00jl01r21jz3hcb1")
   }
 
+    // Useful function to debug the downloaded tiles, can be called from mapScreen if needed
   fun getAllTilesDownloaded() {
     val tileStore = TileStore.create()
     // Get a list of tile regions that are currently available.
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/se/cyrcle/ui/map/MapConfig.kt
```diff
@@ -33,4 +41,82 @@ object MapConfig {
...
+                    zone.boundingBox.southwest(),
+                    zone.boundingBox.northeast(),
+                )))
+
+    // Define style and zooms levels to download
+    val tilesetDescriptor =
+        OfflineManager()
+            .createTilesetDescriptor(
+                TilesetDescriptorOptions.Builder()
+                    .styleURI("mapbox://styles/seanprz/cm27wh9ff00jl01r21jz3hcb1")
```
matttsss, line 82: Maybe we could keep this string in a constant somewhere? I is starting to get duplicated

matttsss (approved):
All good, not much to say except for the small comment and that I am looking forwards for the next PR that will prevent Mapbox from panicking in the Logs when it has no connection xD

Armetiis (approved):
Honestly I have nothing to say on the code. The functionality is also as expected. I just had a weird Toast telling me the download failed (when the logs say otherwise). Nothing too bad, in particular if you're gonna have a PR to show a progression bar where you will dive deeper into thsi

app/src/main/java/com/github/se/cyrcle/ui/map/MapConfig.kt
sean-prz: Done !

Armetiis (approved):
All good!

PR #384, opened 2024-12-08
Enhanced Favorite Parking Cards and Slideable Tab Layout

Armetiis:
### Changes
1. **Redesigned Favorite Parking Cards**
   - Cards are now horizontally elongated for better visual appeal and information display.
   - Each card contains:
     - Parking title
     - Personal note (if available)
     - Icon to remove from favorites (retains existing functionality with confirmation dialog)

2. **Slideable Tab Layout**
   - Users can now switch between "Favorite Parkings" and "My Reviews" tabs by swiping horizontally.
   - This enhances the overall user experience and provides a more intuitive navigation method.

First images shows the new parking cards. Second image showcases the sliding feature.

image
image

### Rationale
The redesigned favorite parking cards focus on displaying the most relevant information to users. The personal note should be sufficient for users to recall why they favorited a particular parking. This approach prevents information overload while still providing key details. Adding more textual information will overload the card, but ... (see Future Improvements)

### Future Improvements
**Additional Parking Information**
   - Once we have appropriate icons, we can consider adding more details such as capacity, rack type, and shelter protection.
   - Using icons instead of text will maintain a clean look while providing more information without overwhelming the user.

### Testing
- Updated existing tests to accommodate the new card layout and slideable tabs.
- Added new tests to ensure proper functionality of the slideable tab layout.

### DIFF
--- a/app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ViewProfileScreenTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ViewProfileScreenTest.kt
@@ -248,14 +248,14 @@ class ViewProfileScreenTest {
 
     composeTestRule.onNodeWithTag("FavoriteParkingList").assertIsDisplayed()
     composeTestRule.onNodeWithTag("ParkingItem0").assertIsDisplayed()
-
+    composeTestRule.onNodeWithTag("ParkingNote0", useUnmergedTree = true).assertIsDisplayed()
     composeTestRule
         .onNodeWithTag("ParkingName0", useUnmergedTree = true)
         .assertIsDisplayed()
         .assertTextContains("Rue de la paix")
 
     composeTestRule.onNodeWithTag("FavoriteParkingList").performScrollToIndex(1)
-
+    composeTestRule.onNodeWithTag("ParkingNote1", useUnmergedTree = true).assertIsDisplayed()
     composeTestRule.onNodeWithTag("ParkingItem1").assertIsDisplayed()
     composeTestRule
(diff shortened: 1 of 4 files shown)


matttsss (commented):
Great feature idea, I just have a little concern with the UI behaviour when there aren't many favorite parkings on the screen. When I swipe left and only have one favorite parking, it teleports to the middle of the column for the animation instead of staying up top. Would it be possible to fix that?

sean-prz (approved):
Looks good and works great, nothing to complain about the code. 
I could ask chatGPT to find me something to complain about or I could be honest and just approve you ! 😉

Armetiis:
> What do you think ?

Honestly I did the change because I also though it was more fluid for the user but this approach has severak drawbacks:

1. Load all the reviews and their corresponding parkings every time user opens the ViewProfileScreen even if they had no intention to open the tab.

2. My solution is not very clean. I think the code looks pretty weird and doesn't really respect the MVVM architecture. However, If I wanted to include this in the ViewModel, it would have added too much unnecessary complications (for this PR).

But honestly I still think it's worth it as it feels much smoother to the user

Armetiis:
> Great feature idea, I just have a little concern with the UI behaviour when there aren't many favorite parkings on the screen. When I swipe left and only have one favorite parking, it teleports to the middle of the column for the animation instead of staying up top. Would it be possible to fix that?

Hi. Sorry I forgot to reply. This was addressed in a recent push. This should be fixed now. I also ensured that the hitbox for the slide is "the whole space" available under the tab manager.

matttsss (approved):
Thank you for the changes, all good for me!

PR #399, opened 2024-12-11
Fix reviews not sorted by time

Pijukai:
## Fix 

When we took a look at the reviews of a parking, the sorting by time was not working properly.  This PR fixes that. 

## How ? 

We needed to sort by seconds and not nanoseconds.


Armetiis (approved):
It works now. Thanks

sean-prz (approved):
Thanks !

PR #400, opened 2024-12-11
Improve Sign-In screen UI

sean-prz:
### Before  vs After
image
image

## How
- Removes create account button instead sign-in with google redirects you if it doesn't find your account
- Transformed the two green buttons into clickable text (with animation on alpha for feedback)
- Moved the buttons down to be closer to the thumb
## Why
Just to make the apps feel less like a student project

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/ui/authentication/SignInScreen.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/authentication/SignInScreen.kt
@@ -29,6 +29,7 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.SpanStyle
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.buildAnnotatedString
+import androidx.compose.ui.text.font.FontStyle
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.style.TextDecoration
 import androidx.compose.ui.text.withStyle
@@ -134,7 +135,9 @@ fun SignInScreen(navigationActions: NavigationActions, userViewModel: UserViewMo
                           }
                     },
                 modifier = Modifier.testTag("AnonymousLoginButton").padding(4.dp),
-                style = TextStyle(textDecoration = TextDecoration.Underline),
+                style =
+                    TextStyle(
+                        textDecoration = TextDecoration.Underline, fontStyle = FontStyle.Italic),
                 onClick = {
                   coroutineScope.launch {
                     clickAnimation(alphaOfGuestButton)
@@ -163,7 +166,9 @@ fun SignInScreen(navigationActions: NavigationActions, userViewModel: UserViewMo
                           }
                     },
                 modifier = Modifier.testTag("OfflineModeButton").padding(4.dp),
-                style = TextStyle(textDecoration = TextDecoration.Underline),
+                style =
+                    TextStyle(
+                        textDecoration = TextDecoration.Underline, fontStyle = FontStyle.Italic),
                 onClick = {
                   coroutineScope.launch {
                     clickAnimation(alphaOfOfflineButton)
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/se/cyrcle/ui/authentication/SignInScreen.kt
```diff
@@ -12,34 +13,40 @@ import androidx.compose.foundation.layout.fillMaxWidth
...
-import com.github.se.cyrcle.ui.theme.ColorLevel
-import com.github.se.cyrcle.ui.theme.atoms.Button
 import com.github.se.cyrcle.ui.theme.atoms.GoogleSignInButton
 import com.github.se.cyrcle.ui.theme.atoms.Text
+import kotlinx.coroutines.launch
 
 @Composable
 fun SignInScreen(navigationActions: NavigationActions, userViewModel: UserViewModel) {
   val context = LocalContext.current
+  val coroutineScope = rememberCoroutineScope()
```
LouisJamesVasseur, line 49: Why does the coroutineScope become necessary here?

LouisJamesVasseur (approved):
After testing everything manually, everything works as advertised. I see no issue with the code even after close inspection, the way you refactored makes sense and it looks good on the Screen. One thing I would say is maybe put the smaller underlined texts in italics? Let me know what you think. I like the way you re-linked the account creation to the Google sign-in, very ingenious! Good work!

sean-prz:
> One thing I would say is maybe put the smaller underlined texts in italics? Let me know what you think. 

Yes I wanted to, but I did not manage to put the text in Italic so I did not waste more time on this, let me try again, I can copy another screen that has italic text

sean-prz:
image
What do you think @LouisJamesVasseur

LouisJamesVasseur:
> image
> What do you think @LouisJamesVasseur

Perfect! Make sure "Continue Offline" is also in italics for consistency!

app/src/main/java/com/github/se/cyrcle/ui/authentication/SignInScreen.kt
```diff
@@ -85,58 +94,90 @@ fun SignInScreen(navigationActions: NavigationActions, userViewModel: UserViewMo
                     when (it) {
                       UserViewModel.SignInFailureReason.ACCOUNT_NOT_FOUND -> {
                         Toast.makeText(context, accountNotFoundToast, Toast.LENGTH_LONG).show()
+                        navigationActions.navigateTo(Screen.CREATE_PROFILE)
```
Pijukai, line 97: As Louis said it is a clean way of refactoring the create account button into the google Sign in one, good and clean job !

app/src/main/java/com/github/se/cyrcle/ui/authentication/SignInScreen.kt
```diff
@@ -85,58 +94,90 @@ fun SignInScreen(navigationActions: NavigationActions, userViewModel: UserViewMo
...
+            val alphaOfOfflineButton = remember { Animatable(1f) }
+            suspend fun clickAnimation(alpha: Animatable<Float, AnimationVector1D>) {
+              alpha.animateTo(
+                  targetValue = 0.1f,
+                  animationSpec =
+                      tween(
+                          durationMillis = 100,
+                      ))
+            }
+            fun resetAlpha(alpha: Animatable<Float, AnimationVector1D>) {
```
Pijukai, line 117: If you are using this function only inside this class maybe you can put it as private ?

sean-prz: This function is already a local function, It can't be used outside of the SignInScreen. The compiler will not even let me write private

Pijukai: Ok it is perfect then !

Pijukai (approved):
The rework works as expected so it is an APPROVED ! Also the style is really great so well done !

PR #402, opened 2024-12-11
Bugfix Initialisation of the List Screen with localisation on

Pijukai:
## Bug

When the localisation permission is granted, on initialisation, the ListScreen displays parking that are far from the current position sometimes more than 1km away despite the maximum radius allowed.

## Fix 

The fix was to make the "MyLocation" suggestion set `cyrcleCenter` to userPosition on Click. Also updating the launchedEffect of the list Screen, passing the argument from `myLocation` to `myLocation.value` contributed to fixing the problem. 
I also updated the documentation.   

This PR completes my first PR #399

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
@@ -509,7 +509,7 @@ class ParkingViewModel(
     // This coroutine will update the closest parkings when all the tiles have been fetched. Note
     // that we don't need to apply the filter again since we use the filteredRectParkings flow.
     viewModelScope.launch {
-      Log.e("ParkingViewModelCenter", "${_circleCenter.value}")
+      Log.d("ParkingViewModelCenter", "${_circleCenter.value}")
       filteredRectParkings
           .map { parkings ->
             parkings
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
```diff
@@ -509,6 +509,7 @@ class ParkingViewModel(
     // This coroutine will update the closest parkings when all the tiles have been fetched. Note
     // that we don't need to apply the filter again since we use the filteredRectParkings flow.
     viewModelScope.launch {
+      Log.e("ParkingViewModelCenter", "${_circleCenter.value}")
```
sean-prz, line 512: Same here, we cam keep it if you want but we should use Log.d rather than Log.d

app/src/main/java/com/github/se/cyrcle/ui/list/ListScreen.kt
```diff
@@ -108,19 +112,23 @@ fun SpotListScreen(
         parking.location.center)
   }
 
-  LaunchedEffect(userPosition, myLocation, chosenLocation.value) {
+  LaunchedEffect(userPosition, myLocation.value, chosenLocation.value) {
+    Log.e("SpotListScreen", "userPosition: $userPosition and chosenLocation: $chosenLocation")
```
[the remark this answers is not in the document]
Pijukai: Yess i will remove them they are not specially useful now!

app/src/main/java/com/github/se/cyrcle/model/parking/ParkingViewModel.kt
Pijukai: Yes this one was and is still useful so I will pass it to Log.d ! Thanks for pointing this out !

sean-prz (approved):
Nice, finally a fully working list screen with search bar, good job.

PR #403, opened 2024-12-11
Let Users see and Delete their images

LouisJamesVasseur:
### What is this PR?

This PR gives more agency to users over the images, that they can now see from the profile and delete

### Why? 

Given that images may contain sensitive content, it's of utmost importance that users are able to have control over the ones they post.

### How?
- New fields for User and Parking now put the Images, both as real data classes and as destinationPaths

Screenshot 2024-12-11 at 18 37 32
Screenshot 2024-12-11 at 18 37 57

- A new column in ViewProfileScreen, where users can see their images as they get added. For each Card inside that Column, the user can either go to that image's Parking or delete it directly
- The buttons displayed when clicking on an image in ParkingDetailsScreen were changed: now, instead of always reporting, it's delete if the user owns the image and report if the user doesn't

### Tests

New tests for the new addImageReport and for the new Composable are provided

### Small comment

Please don't forget to reload the app after deleting an Image from ParkingDetails Screen to see it reflected. The updates in ViewProfileScreen shouldn't require that though


app/src/test/java/com/github/se/cyrcle/model/parking/ParkingViewModelTest.kt
andrepeiry, line 1: Comments in this file are really not useful. `Act`, `Arrange` or `Assert` are easy thing to read from the code if the spacing is done right.
Non-blocking, but keep in mind for the future that comment are here to help understand the code, not blatantly explain the obvious.

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
andrepeiry, line 1: Missing documentation in this file for `loadSelectedUserImages`, `addImageToUserImages` and `removeImageFromUserImages`
[Also `setParkingUser` (even though it's not yours)]

app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ViewProfileScreenTest.kt
andrepeiry, line 1: I think that in general, `assertIsDisplayed` is better than `assertExists`

Also, maybe a test for removing a parking that does not exists or isn't owned by the user would be nice, to reinforce your code.

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
```diff
@@ -455,3 +463,126 @@ private fun UserReviewsSection(
...
+              icon = Icons.Filled.LocalParking,
+              contentDescription = "Go To Parking",
+              testTag = "ParkingFromImageButton",
+              onClick = {
+                parkingViewModel.getParkingById(
+                    showDialogParking.value!!,
+                    { parkingForPath ->
+                      parkingViewModel.selectParking(parkingForPath)
+                      navigationActions.navigateTo(Screen.PARKING_DETAILS)
+                      Toast.makeText(context, strResToast, Toast.LENGTH_SHORT).show()
```
andrepeiry, line 540: I think this toast isn't supposed to be on this button, but rather on the other 😉

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
```diff
@@ -455,3 +463,126 @@ private fun UserReviewsSection(
...
+                    },
+                    {})
+              })
+        },
+        dismissButton = {
+          IconButton(
+              modifier = Modifier.padding(10.dp),
+              icon =
+                  Icons.Filled
+                      .RestoreFromTrash, // or Icons.Filled.Paid or Icons.Filled.AttachMoney etc...
```
andrepeiry, line 550: Not sure I understand the icon choice here. If we want to delete the image from the cloud, then I don't think that an icon named `RestoreFromTrash` is the best choice, and I was confused testing the UI.
Isn't there a simple `Trash` icon ?

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
```diff
@@ -455,3 +463,126 @@ private fun UserReviewsSection(
...
+              onClick = {
+                userViewModel.removeImageFromUserImages(showDialogImagePath.value!!)
+                parkingViewModel.getParkingById(
+                    showDialogParking.value!!,
+                    { parkingForPath ->
+                      parkingViewModel.selectParking(parkingForPath)
+                      parkingViewModel.deleteImageFromParking(
+                          parkingViewModel.selectedParking.value?.uid!!,
+                          showDialogImagePath.value!!)
+                      navigationActions.navigateTo(Screen.VIEW_PROFILE)
```
andrepeiry, line 562: You mentioned that a reboot of the app is needed for the change to take effect. Do you think that some kind of "refresh" function in the view model could be done for this ? I think this might help with the reordering problem found by @sean-prz.

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
LouisJamesVasseur: There unfortunately isn't...

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
LouisJamesVasseur: Mixed things up!

app/src/main/java/com/github/se/cyrcle/model/user/UserViewModel.kt
LouisJamesVasseur: I'll take care of all 4 functions' documentation

andrepeiry (commented):
I'll let you take a look at my inline comments.
I will rewrite one here, as it seems to me that it is quite general to your code, but comment are here to explain design choice, encountered problems and other things, but having
```
// This add one to a
a = a + 1
```
isn't a good comment, as it is easily readable.
It is fine to accompany a block of code with heavy syntax sugar or to separate parts of code.
This is non-blocking, but a good opportunity to drop some knowledge ^^

app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ViewProfileScreenTest.kt
LouisJamesVasseur: I'll make the changes (and great suggestion for the test, thanks!)

app/src/test/java/com/github/se/cyrcle/model/parking/ParkingViewModelTest.kt
LouisJamesVasseur: I'll have another look through them, thanks for bringing this to my attention

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
LouisJamesVasseur: I've addressed all of your comments except this one in my most recent commits. This one's been a long standing issue (not something this PR introduces) that I think warrants its own PR as this is an issue that I've been fighting a lot time now and haven't managed to find a proper fix for quite yet. Does that work for you?

LouisJamesVasseur: EDIT: I managed to fix this, partially. The weird "shuffling" around of images that Sean brought to my attention is fixed, but a reload of the app is still necessary to see the image updates take effect, for a reason still unknown by me for now. I will keep looking.

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
Armetiis: Are you sure? I'm pretty sure it is used in the ZoneManager screen

LouisJamesVasseur: Do you know what it's called then?

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
andrepeiry: I think that's as good as this PR will go ^^ Thank you for doing your best on this !

andrepeiry (approved):

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
LouisJamesVasseur: I just had an idea for this. I'm pretty happy with the fact that the user deleting his Image from the profile screen reflects on the Parking, that's pretty huge for me since that the majority use case. Since it doesn't work on the Parking, though, I'm thinking of changing the DetailsScreen toast to "Image deleted, please reload the app to see change!". What do you think?

sean-prz (approved):
Yeah so after digging about around on main and on your branch, it turns out your PR is affected by a bug that was already existing but just never spotted.

The details screen does become out of sync with the actual data if you reload it.
This means for example when you add a review : 
First you see the new score on the details 
but if you switch to list screen and come back to the details screen 
Then the new score dissapear.

Well it's the same for your picture, If I add image and comeback there a not there and this create all kind of desync bug. 

This bug makes your feature really hard to test as we are limited as only one action before reloading the app, but it seems to work. 

We need to fix this bug as priority 1 next sprint so we can then really test your feature, and have time to address potential bug

LouisJamesVasseur:
> Yeah so after digging about around on main and on your branch, it turns out your PR is affected by a bug that was already existing but just never spotted.
> 
> The details screen does become out of sync with the actual data if you reload it. This means for example when you add a review : First you see the new score on the details but if you switch to list screen and come back to the details screen Then the new score dissapear.
> 
> Well it's the same for your picture, If I add image and comeback there a not there and this create all kind of desync bug.
> 
> This bug makes your feature really hard to test as we are limited as only one action before reloading the app, but it seems to work.
> 
> We need to fix this bug as priority 1 next sprint so we can then really test your feature, and have time to address potential bug
> 
> For the code, I am sorry but as we discussed multiple times : Even if it doesn't make any sense to break down PR we should to do it. I can't review 750 lines of code, I could fake do it with LLM to have a better grade but I don't feel like it

Yep, was made aware of it during testing and did my best to alleviate the bug. Thanks for not holding it against me.

For the number of lines I actually disagree on this one: as you can see from the "Files changed" section, most of those lines is just userViewModels that I had to add everywhere and some extra tests + documentation André (rightfully) asked me to add. My PR before all of those and before tests was about 250 lines of code, and the actual feature I deploy is 1 Composable + its functionality and a 5-line edit of a pre-existing function. This is in-line w/ other PRs on here, and I feel that this is quite an unfair comment for this specific PR.

PR #392, opened 2024-12-09
Feature rewards gambling

AlexandreBeeDeeBop:
This PR introduces the reputation system, in which players can level up and will be able to show off how cool and useful they are to the biker's community !
Users now have a userReputationScore. This score, which we can also also call XP, can be incremented when playing the spin-the-wheel game with their hard earned coins. 
Users can see the progress on the XP bar, and are notified when they level up.
Leveling up gets increasingly hard, except for the legendary segment, which will give you 10 levels no matter what ! 
```
val reputationIncrement = when (segmentName) {
                        rarityLegendary -> 10.0 // Legendary always adds 10.0 level no matter what
                        rarityCommon -> 1.0 / sqrt(currentLevel + 1.0)
                        rarityRare -> 2.0 / sqrt(currentLevel + 1.0)
                        rarityEpic -> 16.0 / sqrt(currentLevel + 1.0)
                        else -> 0.0 // Nothing gives no increment no matter what
                    }
```

Depending on the level, the user will now also have a tag assigned to him next to his username everywhere it appears.
This tag goes like this : 
```
text = "[symbol_of_level level] username"
color = color_of_level
```
Here is how it looks for example at level 52 on the profile screen : 
image
If you want to try all the possible cosmetic tags with their symbols and colors, I recommend you to look on the emulator by running the ViewProfileScreenTest's testInitialDisplayMode() with a Thread.sleep(5000) at the beginning and simply change the userReputationScore of the user in the setup. 
Of course you can also create an account and level up by spending coins on the wheel, but you're going to need a lot of coins and a lot of time...

closes #370

PR #405, opened 2024-12-12
Feature rewards gambling 2

AlexandreMajchrzak:
look at #392 please. I got false banned midway...


app/src/main/java/com/github/se/cyrcle/model/user/User.kt
```diff
@@ -66,3 +66,31 @@ data class User(
...
+          80 to LevelRange("ლ", "#8B0000"), // Dark Red
+          90 to LevelRange("☤", "#B8860B"), // Dark Gold (Goldenrod)
+          100 to LevelRange("♔", "rainbow") // Special case for rainbow effect
+          )
+
+  /** Gets the appropriate level range properties for a given score */
+  fun getLevelRange(score: Double): LevelRange {
+    val level = score.toInt()
+    val rangeStart = ((level / 10) * 10).coerceAtMost(100)
+    return levelRanges[rangeStart] ?: LevelRange("⭒", "#2B2B2B") // Default fallback
```
Armetiis, line 94:
```suggestion
    return levelRanges[rangeStart] ?: levelRanges[0] // Default fallback
```

app/src/main/java/com/github/se/cyrcle/ui/gambling/GamblingScreen.kt
```diff
@@ -281,20 +330,98 @@ fun GamblingScreen(navigationActions: NavigationActions, userViewModel: UserView
...
+                                      .testTag("xp_increment_text"))
+                        }
+                  }
 
               WheelView(
                       modifier = Modifier.size(300.dp).testTag("wheel_view"),
-                      onSegmentLanded = { /* Handle landing */})
+                      onSegmentLanded = { segmentName ->
+                        val reputationIncrement =
+                            when (segmentName) {
```
Armetiis, line 392: Here to avoid "code duplication" you could create a val for the normalization factor `sqrt(currentLevel + 1.0)`

app/src/androidTest/java/com/github/se/cyrcle/ui/gambling/GamblingScreenTest.kt
```diff
@@ -158,7 +199,25 @@ class GamblingScreenTest {
...
-    composeTestRule.mainClock.advanceTimeBy(10000)
+    // Wait for second spin to complete
+    run {
+      val timeoutMs = 15000L
+      val incrementMs = 1000L
+      var elapsedTime = 0L
+
+      while (elapsedTime < timeoutMs) {
+        composeTestRule.mainClock.advanceTimeBy(incrementMs)
+        Thread.sleep(1000)
```
LouisJamesVasseur, line 210: Does the Thread.sleep need to be here?

app/src/androidTest/java/com/github/se/cyrcle/ui/gambling/GamblingScreenTest.kt
LouisJamesVasseur, line 1: In general, there's a lot of constants in this file. Would be nice if you could refactor them into const vals, and remove the Thread.sleeps if at all possible

app/src/main/java/com/github/se/cyrcle/ui/gambling/GamblingScreen.kt
```diff
@@ -181,10 +190,8 @@ fun WheelView(
           rotation = targetRotation % 360
           isSpinning = false
           pauseAfterSpin = true
-          delay(2000)
```
LouisJamesVasseur, line 184: Good work removing the delays!

app/src/main/java/com/github/se/cyrcle/ui/gambling/GamblingScreen.kt
```diff
@@ -209,11 +216,9 @@ fun WheelView(
 
           val startAngle = index * segmentAngle
           val middleAngle = Math.toRadians((startAngle + segmentAngle / 2.0))
-          val textRadius = radius * 0.65f // Moved closer to edge (was 0.6f)
+          val textRadius = radius * 0.65f
```
LouisJamesVasseur, line 219: Maybe a const val here, again?

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
```diff
@@ -460,15 +478,30 @@ fun ParkingDetailsScreen(
...
+                                    text =
+                                        stringResource(
+                                            R.string.display_user_tag_format,
+                                            range.symbol,
+                                            level,
+                                            ownerUsername),
+                                    style = MaterialTheme.typography.bodyMedium,
+                                    color =
+                                        MaterialTheme.colorScheme
+                                            .onSurface // TODO rainbow text color
```
Armetiis, line 492: I see that there are way too many places in the code with an if else on the color argument. Here is my suggestion:
You create a composable in atoms called DisplayText that takes as argument a composable Text and a string representing the color tag. DisplayText will be the one in charge of processing the string input and discriminate betweeen "rainbow" or a hex color tag and this will avoid having and If ("rainbow") ... else ... in all three files that display the username.. I don't know if that was clear. If it wasn't, please let me know. It's ok if for now the color isn't displayed as rainbow. The real problem is that this if else makes the code hard to read and maintain

app/src/androidTest/java/com/github/se/cyrcle/ui/gambling/GamblingScreenTest.kt
AlexandreMajchrzak: yes it does

LouisJamesVasseur: Why is that?

app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ViewProfileScreenTest.kt
```diff
@@ -220,7 +220,7 @@ class ViewProfileScreenTest {
     composeTestRule.onNodeWithTag("EditButton").assertIsDisplayed()
     composeTestRule.onNodeWithTag("DisplayFirstName").assertTextEquals("Jane")
     composeTestRule.onNodeWithTag("DisplayLastName").assertTextEquals("Smith")
-    composeTestRule.onNodeWithTag("DisplayUsername").assertTextEquals("@janesmith")
+    composeTestRule.onNodeWithTag("DisplayUsernameWithLevel").assertTextEquals("[☤92] janesmith")
```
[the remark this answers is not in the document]
AlexandreMajchrzak: These symbols are part of Unicode so there's a good chance that it can display well everywhere. And we're working on an android app anyways for this project so who cares for now.

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
```diff
@@ -80,19 +84,30 @@ fun ParkingDetailsScreen(
...
   val showDialogImage = remember { mutableStateOf<String?>(null) }
   val showDialogImageDestinationPath = remember { mutableStateOf<String?>("") }
+
+  var ownerReputationScore = 0.0
+  var ownerUsername = stringResource(R.string.undefined_username)
   if (selectedParking.owner != "Unknown Owner" && selectedParking.owner != null) {
-    userViewModel.selectSelectedParkingUser(parkingViewModel.selectedParking.value?.owner!!)
-  } else {
-    userViewModel.setParkingUser(null)
+    userViewModel.getUserById(
```
LouisJamesVasseur, line 101: Why did you remove this line? Pretty sure that field needs to be added still

app/src/androidTest/java/com/github/se/cyrcle/ui/gambling/GamblingScreenTest.kt
```diff
@@ -87,16 +89,55 @@ class GamblingScreenTest {
...
     composeTestRule.onNodeWithTag("coin_display").assertTextEquals("Coins: 25")
+
     // First spin (25 coins -> 15 coins)
     composeTestRule.onNodeWithTag("spin_button").assertIsEnabled()
     composeTestRule.onNodeWithTag("spin_button").performClick()
-    composeTestRule.mainClock.advanceTimeBy(10000)
+
+    // Wait for first spin to complete
+    run {
+      val timeoutMs = 15000L
```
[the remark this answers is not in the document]
AlexandreMajchrzak: the wheel being in constant movement, the testing is absolutely horrendous and i have to pause the clock and advance time manually, and after trial and error i came up with all these values you can see in here. A spin wont take more than 15 seconds.

Armetiis (changes requested):
Hi. I think you did a pretty good job overrall. However, for some reason i still can't see the level of other users (e.g. in the image below, gokunaruto... should be level 15 but I see level 0).

Also please mind my comment on the duplication of if (color == "rainbow") ... else ...
These are pretty big pieces of code that are duplicated (three times). I'd be glad to explain further the implementation i have in mind if my comment was not clear enough. 

image

app/src/main/java/com/github/se/cyrcle/ui/gambling/GamblingScreen.kt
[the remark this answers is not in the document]
Armetiis: I agree with Louis here!!

app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ViewProfileScreenTest.kt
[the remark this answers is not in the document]
Armetiis: I think the choice of Unicode characters was very thoughtful and very minecraft factions of you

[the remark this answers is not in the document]
LouisJamesVasseur: Agreed. Good job making sure that this was taken into account, shows you put some good thought into it!!

app/src/androidTest/java/com/github/se/cyrcle/ui/gambling/GamblingScreenTest.kt
AlexandreMajchrzak: Good question, i dont have the answer but the test doesnt pass without, or with a lower value...

app/src/androidTest/java/com/github/se/cyrcle/ui/profile/ViewProfileScreenTest.kt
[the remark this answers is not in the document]
AlexandreMajchrzak: > I think the choice of Unicode characters was very thoughtful and very minecraft factions of you. SPEAKING OF FACTIONS (just kidding)

I literally took minecraft inspiration. I was on a bunch of lifetime leaderboards on hypixel for years ;) haha

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
LouisJamesVasseur: Yeah, I was also thinking something along those lines but you beat me to it. It's a Composable that's often repeated in code so it makes sense to use an Atom for this, w/ the only "varying" factor being the color

app/src/main/java/com/github/se/cyrcle/model/user/User.kt
AlexandreMajchrzak: you're right, that's cleaner

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
```diff
@@ -449,7 +449,8 @@ private fun UserReviewsSection(
                       },
                       options = mapOf(),
                       userViewModel = userViewModel,
-                      reviewViewModel = reviewViewModel)
+                      reviewViewModel = reviewViewModel,
+                      ownerReputationScore = null)
```
LouisJamesVasseur, line 453: Why put the ownerReputationScore as null and not simply 0 here?

LouisJamesVasseur (commented):
Already mentioned how impressed I was in general w/ the feature, and I'm happy to see that some of my feedback was taken into account already. Please mind the comments I added below

app/src/main/java/com/github/se/cyrcle/model/user/User.kt
AlexandreMajchrzak: except that it doesnt work ??

app/src/main/java/com/github/se/cyrcle/ui/gambling/GamblingScreen.kt
AlexandreMajchrzak: indeed

app/src/main/java/com/github/se/cyrcle/model/user/User.kt
AlexandreMajchrzak: because if i take your suggestion the return type needs to be changed to LevelRange? and so now everywhere i call the function i need to add safe "?" call

app/src/main/java/com/github/se/cyrcle/ui/gambling/GamblingScreen.kt
AlexandreMajchrzak: yes why not

app/src/main/java/com/github/se/cyrcle/ui/gambling/GamblingScreen.kt
```diff
@@ -281,20 +330,98 @@ fun GamblingScreen(navigationActions: NavigationActions, userViewModel: UserView
...
+                          Text(
+                              text = stringResource(R.string.level_display, (displayedLevel + 1)),
+                              style = MaterialTheme.typography.bodyMedium,
+                              modifier = Modifier.testTag("next_level_text"))
+                        }
+
+                    LinearProgressIndicator(
+                        progress = { animatedProgress.value },
+                        modifier =
+                            Modifier.fillMaxWidth(0.8f).height(8.dp).testTag("level_progress_bar"),
```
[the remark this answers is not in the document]
AlexandreMajchrzak: by camel case you mean LevelProgressBar instead of snake level_progress_bar ?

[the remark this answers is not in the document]
LouisJamesVasseur: Exactly right!

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
AlexandreMajchrzak: this was a suggestion from Maria, because in the ViewProfileScreen we dont see the owner of the review but the name of the parking being reviewed. So I call the same previewcard but with this argument as null to indicate to the function PreviewCard that it's a call from ViewProfileScreen and that i dont need to see the owner's name with his tag and stuff...

app/src/main/java/com/github/se/cyrcle/ui/gambling/GamblingScreen.kt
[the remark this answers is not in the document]
AlexandreMajchrzak: i changed them all one by one in GamblingScreen and in GamblingScreenTest. Took me quite a while because chatgpt is too dumb to do that repetitive task... And i thought AI was gonna steal my job...

app/src/main/java/com/github/se/cyrcle/ui/profile/ViewProfileScreen.kt
LouisJamesVasseur: Ok, fair enough!

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
AlexandreMajchrzak: Ok it's a bit too complicated for no reason to do that. Because the texts are from different instances (whether material3 or atoms or idk what else  even...), so their arguments differ. The parameters are different in each file.
And instead I ask you to accept my counter-offer, which is that i place the if else condition inside the color argument itself, reducing code size by 2 everywhere. That's already good enough.

LouisJamesVasseur: Unfortunate that you couldn't get our suggestion to work, but at the same time I understand that it's difficult and it's not really this PR's objective anyway. I'm okay w/ your suggestion, how about you @Armetiis?

AlexandreMajchrzak:
> Hi. I think you did a pretty good job overrall. However, for some reason i still can't see the level of other users (e.g. in the image below, gokunaruto... should be level 15 but I see level 0).
> 
> Also please mind my comment on the duplication of if (color == "rainbow") ... else ... These are pretty big pieces of code that are duplicated (three times). I'd be glad to explain further the implementation i have in mind if my comment was not clear enough.
> 
> image

honestly i cant find anything suspicious that could explain that in my AllReviewsScreen code. But I have an issue with a test in MainActivityTest and it's the getUserById function that is not working. So maybe that's the root of the problem since i call it to get the users' reputations.

LouisJamesVasseur:
> > Hi. I think you did a pretty good job overrall. However, for some reason i still can't see the level of other users (e.g. in the image below, gokunaruto... should be level 15 but I see level 0).
> > Also please mind my comment on the duplication of if (color == "rainbow") ... else ... These are pretty big pieces of code that are duplicated (three times). I'd be glad to explain further the implementation i have in mind if my comment was not clear enough.
> > image
> 
> honestly i cant find anything suspicious that could explain that in my AllReviewsScreen code. But I have an issue with a test in MainActivityTest and it's the getUserById function that is not working. So maybe that's the root of the problem since i call it to get the users' reputations.

Might be able to help. Could you elaborate on what you mean by "not working"?

AlexandreMajchrzak:
> > > Hi. I think you did a pretty good job overrall. However, for some reason i still can't see the level of other users (e.g. in the image below, gokunaruto... should be level 15 but I see level 0).
> > > Also please mind my comment on the duplication of if (color == "rainbow") ... else ... These are pretty big pieces of code that are duplicated (three times). I'd be glad to explain further the implementation i have in mind if my comment was not clear enough.
> > > image
> > 
> > 
> > honestly i cant find anything suspicious that could explain that in my AllReviewsScreen code. But I have an issue with a test in MainActivityTest and it's the getUserById function that is not working. So maybe that's the root of the problem since i call it to get the users' reputations.
> 
> Might be able to help. Could you elaborate on what you mean by "not working"?

look at the CI, there's this test that fails to onSuccess when calling getUserById

app/src/main/java/com/github/se/cyrcle/ui/gambling/GamblingScreen.kt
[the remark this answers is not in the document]
andrepeiry: The reason that your CI does not pass is for this : 
```
override fun getUserById(
      userId: String, onSuccess: (User) -> Unit, onFailure: (Exception) -> Unit
  ) {
    db.collection(collectionPath).document(userId).get()
        .addOnSuccessListener { document ->
          val userPublic: UserPublic =
              try {
                Log.d("UserRepositoryFirestore", " Correctly fetched UserPublic")
                deserializeUserPublic(document.data!!)
              } catch (e: Exception) {
                Log.d("UserRepositoryFirestore", "Error fetching UserPublic")
                onFailure(e)                                                          // onFailure
                return@addOnSuccessListener
              }
          db.collection(collectionPath)
              .document(userId).collection("private").document("details").get()
              .addOnSuccessListener { detailsDocument ->
                val userDetails = deserializeUserDetails(detailsDocument.data!!)
                onSuccess(User(userPublic, userDetails))                              // onSuccess
              }
              .addOnFailureListener { onSuccess(User(userPublic, null)) }             // onSuccess
        }
        .addOnFailureListener(onFailure)                                              // onFailure
  }
```
Somehow, your modification in `deserializeUserPublic` does not have the same behavior as before. Maybe try another way to deal with `userReputationScore` that could avoid an error down the line. It is calling `onFailure` instead of `onSuccess`, even though you seems to obtain the good return value.

app/src/main/java/com/github/se/cyrcle/model/user/UserRepositoryFirestore.kt
```diff
@@ -151,7 +151,11 @@ constructor(private val db: FirebaseFirestore, private val auth: FirebaseAuth) :
 
   private fun deserializeUserPublic(data: Map<String, Any>): UserPublic {
     val gson = Gson()
-    return gson.fromJson(gson.toJson(data), UserPublic::class.java)
+    val json = gson.toJson(data)
+    val userPublic = gson.fromJson(json, UserPublic::class.java)
+    // Ensure default values for missing fields
+    val userReputationScore = (data["userReputationScore"] as? Number)?.toDouble() ?: 0.0
+    return userPublic.copy(userReputationScore = userReputationScore)
```
andrepeiry, line 158: The reason that your CI does not pass is for this : 
```
override fun getUserById(
      userId: String, onSuccess: (User) -> Unit, onFailure: (Exception) -> Unit
  ) {
    db.collection(collectionPath).document(userId).get()
        .addOnSuccessListener { document ->
          val userPublic: UserPublic =
              try {
                Log.d("UserRepositoryFirestore", " Correctly fetched UserPublic")
                deserializeUserPublic(document.data!!)
              } catch (e: Exception) {
                Log.d("UserRepositoryFirestore", "Error fetching UserPublic")
                onFailure(e)                                                          // onFailure
                return@addOnSuccessListener
              }
          db.collection(collectionPath)
              .document(userId).collection("private").document("details").get()
              .addOnSuccessListener { detailsDocument ->
                val userDetails = deserializeUserDetails(detailsDocument.data!!)
                onSuccess(User(userPublic, userDetails))                              // onSuccess
              }
              .addOnFailureListener { onSuccess(User(userPublic, null)) }             // onSuccess
        }
        .addOnFailureListener(onFailure)                                              // onFailure
  }
```
Somehow, your modification in `deserializeUserPublic` does not have the same behavior as before. Maybe try another way to deal with `userReputationScore` that could avoid an error down the line. It is calling `onFailure` instead of `onSuccess`, even though you seems to obtain the good return value.

LouisJamesVasseur: Yes, I can confirm this from looking at the call stack of the failing test. I've been looking into why, still at it...

LouisJamesVasseur: Well it seems from your new push that the old serializer works for this. I have no objections to reverting as we'll remove the default values for null fields from the deserializers when we clean the database next week, so good work finding a fix!

AlexandreMajchrzak:
please god, make it stop

app/src/main/java/com/github/se/cyrcle/ui/gambling/GamblingScreen.kt
```diff
@@ -35,6 +44,60 @@ import com.github.se.cyrcle.ui.theme.molecules.TopAppBar
...
+ * 5. Text Positioning:
+ *     - Radial text placement using polar coordinates
+ *     - Text radius: 65% of wheel radius (textProportion = 0.65f)
+ *     - Text rotation: Compensates for segment angle + 90° for readability
+ *     - Dynamic scaling: Font size = 6% of wheel radius
+ *
+ * @param modifier Modifier for the composable
+ * @param onSegmentLanded Callback triggered when wheel stops on a segment
+ * @return A function that triggers the wheel spin
+ */
```
Armetiis, line 100: Thanks for documenting this!!

app/src/androidTest/java/com/github/se/cyrcle/ui/gambling/GamblingScreenTest.kt
LouisJamesVasseur: Ok got it, I'll let it go as-is then.

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
Armetiis: I think we can keep this for now. I suggested that because I was unsure of how complicated it would actually be to make the text rainbow colored. I doubt it's just a "Color" argument. But this is not a concern for now. This can be addressed when actually implementing the rainbow color, as that's when it will be possible to assess if it's necessary or not.

For the moment this is perfect, as the main concern to me was the heavy code duplication.

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
LouisJamesVasseur: Nice, you changed it!

LouisJamesVasseur (approved):
You've addressed all the changes I requested and I've found no more bugs. Your PR is approved from me, good work!

Armetiis (approved):
Good job on fixing everything! I don't think there's much left to say. Approved

app/src/main/java/com/github/se/cyrcle/ui/gambling/GamblingScreen.kt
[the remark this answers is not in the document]
AlexandreMajchrzak: Here you go, MASSIVE documentation (and you know what else is massive ? the low taper fade meme)

app/src/main/java/com/github/se/cyrcle/ui/parkingDetails/ParkingDetailsScreen.kt
AlexandreMajchrzak: I deleted that during a rebase when I shouldnt have.

app/src/main/java/com/github/se/cyrcle/ui/gambling/GamblingScreen.kt
AlexandreMajchrzak: You're welcome

PR #408, opened 2024-12-13
Tutorial screen

### LINKED ISSUE #397: Tutorial Screen

andrepeiry:
# A small tutorial to teach new user how the small tricks of the app ;)

## New screen

After creating an account, the user is redirect to the tutorial (instead of the map before). There he will see the following
Tutorial Welcome Tutorial for Map Tutorial for List Tutorial Thank You

A button was added in the Profile screen to allow a user to view the tutorial again.

## Technicalities

There are no MVVM model, as this would be more cumbersome that helpful in this case.

Route were changed 
- At the end of the account creation, the user is directed to the tutorial, rather the map
- At the end of the tutorial, the user is sent to the map
- The user can skip the tutorial with the appropriate button to directly go to the map
  - If the user comes from the Profile screen, it will be redirected there instead

## Close

Close #397


AlexandreBeeDeeBop:
Quick question : Can the user watch the tutorial again later ? Like is there a button to watch it again ? Maybe just in case they skipped but regret it or if they just forgot how to do something.

andrepeiry:
> Quick question : Can the user watch the tutorial again later ? Like is there a button to watch it again ? Maybe just in case they skipped but regret it or if they just forgot how to do something.

@AlexandreBeeDeeBop This has been addressed in the latest commits.

LouisJamesVasseur (commented):
Screenshot 2024-12-19 at 18 11 29

Little visual bug on my Android ;)

app/src/androidTest/java/com/github/se/cyrcle/ui/profile/CreateProfileScreenTest.kt
```diff
@@ -70,7 +70,7 @@ class CreateProfileScreenTest {
 
     composeTestRule.onNodeWithTag("AuthenticateButton", useUnmergedTree = true).performClick()
 
-    verify(mockNavigationActions).navigateTo(TopLevelDestinations.MAP)
+    verify(mockNavigationActions).navigateTo(Screen.TUTORIAL)
```
LouisJamesVasseur, line 73: Making sure I get it: the MockNavigationActions here now goes to the tutorial on first sign in instead of the map, I get that. But I don't get why you straight up removed the test. Is there no way to simulate the already signed-in version of this, which should still lead to the map? Is this test replaced by anything somewhere else?

app/src/main/java/com/github/se/cyrcle/ui/tutorial/TutorialScreen.kt
```diff
@@ -0,0 +1,202 @@
...
+  Spacer(Modifier.height(subtitleSpacer))
+
+  Text(
+      text = stringResource(R.string.tutorial_screen_thank_you_subtitle),
+      style = MaterialTheme.typography.bodyMedium,
+      modifier = Modifier.padding(generalPadding),
+      testTag = "TutorialScreenThankYouSubtitle")
+}
+
+// Uncomment to have a preview of the Tutorial
```
LouisJamesVasseur, line 197: It says uncomment in this comment, but in your PR it's uncommented already. Normal?

LouisJamesVasseur (commented):
Good work! Please mind my inline comments. 

UI-wise, I'd like to see a back button added next to the "next". This is non-blocking but I think it could be cool

app/src/androidTest/java/com/github/se/cyrcle/ui/profile/CreateProfileScreenTest.kt
andrepeiry: What you call "first sign in" in in reality "create account", which you can only do once per Google account, so the next time, when the user sign in, it goes from *Sign In* to *Map*, whereas this test is checking *CreateProfile* to *Tutorial*

LouisJamesVasseur: Thanks for clarifying. As per my original question, was the original test straight-up removed?

app/src/main/java/com/github/se/cyrcle/ui/tutorial/TutorialScreen.kt
andrepeiry: I think it was meant to be commented regarding the code coverage
I left it because it is hard to see the tutorial from a phone in which you've already have an account (before the button in profile).
As there is now a button, I will remove this.

LouisJamesVasseur:
> > UI-wise, I'd like to see a back button added next to the "next". This is non-blocking but I think it could be cool
> 
> I think that for a 4-screen tutorial, we can "afford" the user to missclick, as there is now a button to watch the tutorial again.

This is a fair point, I'll accept it

matttsss (commented):
Very nice, the UI explanations are good and well presented, I am just a bit confused with the navigation if you could help me out on that?

app/src/androidTest/java/com/github/se/cyrcle/ui/profile/CreateProfileScreenTest.kt
andrepeiry: No test was removed, only the destination was changed.

LouisJamesVasseur: Understood!

andrepeiry:
@matttsss Thank you for pointing the navigation problems, it was a big misunderstanding / bad design on my part. It should be fixed now ^^

matttsss (approved):
All good for me, thanks for the changes on the navigation!

LouisJamesVasseur (approved):
As the last approver on the last PR (I think?) of the project I have one thing to say: great work!

LouisJamesVasseur (approved):

Issue #381, opened 2024-12-08
User should see the progress and size of their download


sean-prz:
Currently no informations is displayed to the user about the status of the download, he doesn't know when the. map is done downloading or the size of the download

sean-prz:
We will not address size as we're running out of time

PR #415, opened 2024-12-13
Progress bar for download

sean-prz:
- fix : #381 
## What : 
Adds a progress bar when downloading zones for offline
image

## How
- Nothing is change behind the scene, there is still no option to cancel the download or no edge case handling if you loose connection, because we're out of time. 
- This just adds a feedback to show the download progress

## Why
it felt strange to click download and not have any feedback.

### DIFF
--- a/app/src/androidTest/java/com/github/se/cyrcle/ui/zone/ZoneSelectionScreenTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/ui/zone/ZoneSelectionScreenTest.kt
@@ -59,7 +59,7 @@ class ZoneSelectionScreenTest {
   @Test
   fun checkAllUIElementsAreDisplayed() {
     composeTestRule.setContent {
-      ZoneSelectionScreen(navigationActions, mapViewModel, parkingViewModel, addressViewModel)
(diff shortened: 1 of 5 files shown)


matttsss (approved):
Not much to say...  Maybe missing a test but I could understand that it is hard to test that

Pijukai:
I don't have much to say either despite that the fact a more complete would be welcome. However just before approving, I have a minor suggestion. For me instead of hide the button below the download bar should be cancel and we should be able to make the dialog screen hide when clicking anywhere else except on the dialog screen. It is however just a suggestion and If it is too difficult or time consuming just tell me.

sean-prz:
@Pijukai, I could cancel the download, but it would be quite a big feature. 
I would need to cancel the download of mapbox then cancel the download of the parkings, and finally clear everything and remove the zone. 
It would make a big PR, and since we're in the polish sprint I decided to go with just hide. 

I don't want them to be able to click on the side of the alertDialog, because hiding the download navigates you to another screen and it would be non intuitive if clicking on the side of the dialog navigated you to another screen.

sean-prz:
Thank you !

Issue #411, opened 2024-12-13
New parkings don't have any minimum size


sean-prz:
## Problem 
User can create parkings with almost null area, making them disappear on the advanced mode.

PR #418, opened 2024-12-13
Assert a minimum area for new parking

sean-prz:
- fix : #411 
## What
Ensure parking are minimum 5 meters square

## How
reuse the `isValid` function that already asserted an upper bound
adapts test so that they don't create a parking with a 0 area anymore

## Why
to prevent having invisible parking on the advanced mode, because rectangles are too small to be visible.

### DIFF
--- a/app/src/androidTest/java/com/github/se/cyrcle/ui/testEnd2End/MainActivityTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/ui/testEnd2End/MainActivityTest.kt
@@ -21,6 +21,7 @@ import androidx.compose.ui.test.performTextInput
 import androidx.compose.ui.test.performTouchInput
 import androidx.compose.ui.test.swipe
 import androidx.compose.ui.test.swipeRight
+import androidx.compose.ui.test.swipeUp
 import com.github.se.cyrcle.MainActivity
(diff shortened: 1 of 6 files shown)


LouisJamesVasseur (commented):
Cool, works as intended, good work.
I'd like to see the toast that's displayed when the Parking is too small changed. "Please select a valid area" isn't super for users (especially first-time ones) to understand what they did wrong. "Please make your Parking larger" or something like that would be more adequate to the feature you're implementing

Armetiis (approved):
Clean reuse of code. Works as intended

sean-prz:
@LouisJamesVasseur 
 The function of the viewmodel that checks if the area is valid doesn't return the informations of why the area is invalid to the UI. I don't want to risk modifying it so late in the project, so that I can make the toast dynamically display why the area is invalid even if I agree this is a good idea. 
 
 As a workaround can I maybe change the text toast to inform the user what we qualify as a valid area ? (the size it should be) .
Do you have any other idea ?

LouisJamesVasseur:
> @LouisJamesVasseur, I changed the toast to this, what do you think? image

Yep that works for me.

LouisJamesVasseur (approved):
Thanks for changing the Toast! Much more user-friendly now!

Issue #383, opened 2024-12-08
Navigate from zone management to map 


sean-prz:
### Problem

User should be able to navigate from the zone selection screen to the map, with the map centered on the zone clicked. 
Without this feature, the only way to go to a zone on the map would be for the user to scroll there, because the search bar doesn't work offline. 

### Solution
Re-use function that will soon be introduced by @andrepeiry that allows to center the map on a location.
Could also add the zones in the search bar offline to make it usefull.

PR #419, opened 2024-12-13
Go to map from zone manager screen

sean-prz:
- fix : #383 
## What
Allow user to open the map centered on a zone he downloaded

image

## How 
Re-using André's function for the go-to-map feature but overloading it with a zoom parameter, because we don't want to be too zoomed here

## Why
Because offline, the search bar doesn't work and so navigating between areas is only possible via scrolling


matttsss:
Everything look in order. But please remember to test your feature 😅

andrepeiry:
> @andrepeiry Can you review pls

Sorry, completely forgot about this PR, on it right now !

andrepeiry (commented):
Everything seems in order, but as @matttsss stated, a small test for the navigation (`verifiy` that clicking on the button calls navigate to the desired destination would be nice ^^

sean-prz:
@matttsss @andrepeiry 
On your request, I added a small test that ensure my new feature is working, check out the latest commit ! 👍

andrepeiry (approved):
Everything seems nice to me, you're good to go ! Sorry again for the delay ...

matttsss (approved):
All good for me, thanks for the changes!

PR #420, opened 2024-12-13
LET IT SNOW ❄️

Armetiis:
This PR introduces a delightful snowfall animation to our app, enhancing user engagement during the holiday season. Users can now shake their devices to start and stop the snowfall, creating a festive atmosphere.

### Snowfall Animation:
- Implemented a snowflake animation using Jetpack Compose.
- Snowflakes fall from the top of the screen at random positions.
- Each snowflake has a different size and speed, making the result feel very natural
- Added a "blizzard" background, which is just a shade of Cerulean with a very low alpha. This is needed to see the snowflakes given that our app is mainly white

### Shake Detection Implementation:
- Implemented shake detection to allow users to trigger the snowfall with device shakes.
- Added a 2-second cooldown period after starting or stopping the snowfall to prevent accidental toggling.

-- -
It's not very easy to see on a screenshot, but try it and you'll see!

image
image

**Note:**
It is still possible to interact with the app while the snowflakes are falling, ensuring that users can continue their activities without interruption.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -8,7 +8,7 @@
     <uses-permission android:name="android.permission.READ_MEDIA_IMAGES"/>
 
     <!-- Permission to access the sensor of the device -->
-    <uses-feature android:name="android.hardware.sensor.accelerometer" android:required="true" />
+    <uses-feature android:name="android.hardware.sensor.accelerometer"/>
 
 
     <application
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -5,7 +5,10 @@
     <!-- Permission of MapBox to access Coarse Location of the user -->
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
-   <uses-permission android:name="android.permission.READ_MEDIA_IMAGES"/>
+    <uses-permission android:name="android.permission.READ_MEDIA_IMAGES"/>
+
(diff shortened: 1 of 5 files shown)


sean-prz (commented):
Pretty !!
Do you think you can move some of the code outside the Mainactivity to not pollute it too much ?

Armetiis:
> Do you think you can move some of the code outside the Mainactivity to not pollute it too much ?

Thank you for the feedback. I have created a class `ShakeDetector` that inherits from `SensorEventListener ` and I have to admit it is much cleaner. I also created a helper function to show the Toast. I don't think I could make it better. Let me know if you have further suggestions

app/src/main/AndroidManifest.xml
```diff
@@ -5,7 +5,10 @@
     <!-- Permission of MapBox to access Coarse Location of the user -->
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
-   <uses-permission android:name="android.permission.READ_MEDIA_IMAGES"/>
+    <uses-permission android:name="android.permission.READ_MEDIA_IMAGES"/>
+
+    <!-- Permission to access the sensor of the device -->
+    <uses-feature android:name="android.hardware.sensor.accelerometer" android:required="true" />
```
sean-prz, line 11: it seems that the required:true is meant to be used if the permission is necessary for the app to work correctly, and I don't think it's the case here.

We can remove it and the app should still work.

matttsss (approved):
Not much to say, a neat integration with the MainActivity

sean-prz (approved):
👍

PR #423, opened 2024-12-16
Bugfix topappbar title overflows

Armetiis:
### Description:

This PR addresses the issue of TopAppBar title overflowing beyond two lines, improving readability and UI consistency across the app.

### Changes:
**1. Implemented AutoResizingText composable:**
- Created a new `AutoResizingText` composable that dynamically adjusts font size to fit within the available space.
- Both font size and line height are reduced by 5% in each iteration until text fits or reaches a minimum font size of 6sp to maintain proper text scaling.
- Replaced the standard `Text` composable with the new `AutoResizingText` in the TopAppBar with a maximum of 2 lines

**2. Improved UI layout:**
- Added horizontal padding to the title for better spacing.
- Adjusted the navigation icon to use a semi-transparent white background for improved visibility.

### Before vs after:

Let us see the difference between before and after for the same "long title"

image
image
image

### DIFF
--- a/app/src/androidTest/java/com/github/se/cyrcle/ui/theme/molecules/TopAppBarTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/ui/theme/molecules/TopAppBarTest.kt
@@ -76,4 +76,28 @@ class TopAppBarTest {
         .assertIsDisplayed()
         .assertHasClickAction()
   }
+
+  @Test
+  fun displayTopAppBarWithLongTitle() {
+    val tag2 = "TopAppBar2"
+    val title =
+        "This is a very long title that should be truncated but just in case it is not, I am making it longer"
+    composeTestRule.setContent {
+      Scaffold(topBar = { TopAppBar(navigationActions, title, tag2) }) { innerPadding ->
+        Column(modifier = Modifier.padding(innerPadding)) {}
+      }
+    }
+
+    // Top App Bar
+    composeTestRule.onNodeWithTag(tag2).assertIsDisplayed()
+
+    // Title
+    composeTestRule.onNodeWithTag("${tag2}Title", true).assertIsDisplayed().assertTextEquals(title)
+
+    // Ensure button is still displayed
+    composeTestRule
+        .onNodeWithTag("${tag2}GoBackButton", true)
+        .assertIsDisplayed()
+        .assertHasClickAction()
+  }
 }
--- a/app/src/main/java/com/github/se/cyrcle/ui/theme/molecules/TopAppBar.kt
+++ b/app/src/main/java/com/github/se/cyrcle/ui/theme/molecules/TopAppBar.kt
@@ -1,5 +1,10 @@
 package com.github.se.cyrcle.ui.theme.molecules
 
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.automirrored.filled.ArrowBack
 import androidx.compose.material3.CenterAlignedTopAppBar
@@ -14,6 +19,9 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.LocalConfiguration
 import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.style.TextOverflow
+import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import com.github.se.cyrcle.ui.navigation.NavigationActions
 import com.github.se.cyrcle.ui.theme.Typography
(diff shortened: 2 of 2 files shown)


app/src/androidTest/java/com/github/se/cyrcle/ui/theme/molecules/TopAppBarTest.kt
```diff
@@ -76,4 +76,28 @@ class TopAppBarTest {
...
+      Scaffold(topBar = { TopAppBar(navigationActions, title, tag2) }) { innerPadding ->
+        Column(modifier = Modifier.padding(innerPadding)) {}
+      }
+    }
+
+    // Top App Bar
+    composeTestRule.onNodeWithTag(tag2).assertIsDisplayed()
+
+    // Title
+    composeTestRule.onNodeWithTag("${tag2}Title", true).assertIsDisplayed().assertTextEquals(title)
```
Pijukai, line 95: Just a little clarification here, the test is supposed to verify that the title was truncated but here it tests that the topAppBar title is the original non truncated. Am I missing something ?

app/src/main/java/com/github/se/cyrcle/ui/theme/molecules/TopAppBar.kt
```diff
@@ -49,10 +61,57 @@ fun TopAppBar(navigationActions: NavigationActions, title: String, testTag: Stri
...
+ * will be resized until it fits the available space, or until the font size is 6sp. Both the font
+ * size and the line height will be reduced by 5% at each iteration.
+ *
+ * @param text The text to display
+ * @param initialFontSize The initial font size of the text
+ * @param modifier The modifier of the text
+ * @param maxLines The maximum number of lines the text can take
+ */
+@Composable
+fun AutoResizingText(
```
Pijukai, line 91: Modularization makes the code a lot clearer well done !

app/src/main/java/com/github/se/cyrcle/ui/theme/molecules/TopAppBar.kt
```diff
@@ -49,10 +61,57 @@ fun TopAppBar(navigationActions: NavigationActions, title: String, testTag: Stri
...
+          IconButton(
+              onClick = { navigationActions.goBack() },
+              modifier =
+                  Modifier.testTag("${testTag}GoBackButton")
+                      .background(Color.White.copy(alpha = 0.1f), shape = RoundedCornerShape(12.dp))
+                      .size(40.dp)) {
+                Icon(
+                    Icons.AutoMirrored.Filled.ArrowBack,
+                    contentDescription = "Go Back",
+                    modifier = Modifier.size(24.dp))
```
Pijukai, line 74: I don't understand why here the `IconButton` contains an `Icon` since the `IconButton` can do everything the `Icon` does. Is it a question of size management ?

app/src/main/java/com/github/se/cyrcle/ui/theme/molecules/TopAppBar.kt
```diff
@@ -49,10 +61,57 @@ fun TopAppBar(navigationActions: NavigationActions, title: String, testTag: Stri
...
+      color = Color.White,
+      fontSize = fontSize,
+      fontWeight = FontWeight.Bold,
+      lineHeight = lineHeight,
+      maxLines = maxLines,
+      modifier = modifier.drawWithContent { if (readyToDraw) drawContent() },
+      onTextLayout = { textLayoutResult ->
+        if (textLayoutResult.didOverflowHeight && fontSize > 6.sp) {
+          fontSize *= 0.95f
+          lineHeight *= 0.95f
```
Pijukai, line 112: Just for those two values did you consider a constant value for 0.95f if they are both related of course.

app/src/androidTest/java/com/github/se/cyrcle/ui/theme/molecules/TopAppBarTest.kt
Armetiis: The whole point of the feature is not to truncate the title. I think what mislead you is the content of the title but consider it as "would be truncated without this feature" rather than "should be truncated"

app/src/main/java/com/github/se/cyrcle/ui/theme/molecules/TopAppBar.kt
Armetiis: Thank you!

app/src/main/java/com/github/se/cyrcle/ui/theme/molecules/TopAppBar.kt
Armetiis: This is just how material3 IconButton composables are. I don't make the rules

app/src/androidTest/java/com/github/se/cyrcle/ui/theme/molecules/TopAppBarTest.kt
Pijukai: Ohhhh makes sense thanks ! It might have something to do with the title yes.

app/src/main/java/com/github/se/cyrcle/ui/theme/molecules/TopAppBar.kt
Pijukai: Icon button has a modifier and a vector Image which makes the Icon useless. I think the Icon here is not necessary unless it is for size management.

Armetiis: I think you're talking about the IconButtom from the atoms. It doesn't work for me. I need to use material3's IconButtom

Pijukai: Yes, It was the one I was talking about but if it doesn't work then you made the right choice I won't block your PR for this.

Pijukai (approved):
Congrats for this PR you made a great job! I tried for the IconButton but i wasn't able to make the atom works either so it is non blocking for me. APPROVED !

app/src/main/java/com/github/se/cyrcle/ui/theme/molecules/TopAppBar.kt
Armetiis: Btw just to clarify when I say it doesn't work i meant that it doesn't suit my needs. Not that it actually doesn't work

sean-prz (approved):
Gtg ( it means good to go )

Issue #386, opened 2024-12-08
Update MapScreen on network changes


matttsss:
When the network connectivity is changed, the Map is not updated. A possible solution would be to add a launched effect to detect such changes.

PR #425, opened 2024-12-16
Bugfix MapScreen network changes

matttsss:
## What is this PR? 
This PR aims to bugfix the Network connectivity changes not updating the parkings on the map.

## How?
This was done by adding a state in the `MapViewModel` that remembers the screen location coordinates. This allows to add a `LaunchedEffect` in the `MapScreen` that updates on connectivity changes.

## Testing:
To test that the behavior is correct: swapping between offline and online mode should keep the parkings displayed without having to move around the map.

* closes: #386

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
@@ -52,6 +52,11 @@ class MapViewModel : ViewModel() {
           } ?: false
         }
 
+  private val _mapScreenCoordinates =
+      MutableStateFlow<Pair<Point, Point>>(
+          Pair(Point.fromLngLat(0.0, 0.0), Point.fromLngLat(0.0, 0.0)))
+  val mapScreenCoordinates: StateFlow<Pair<Point, Point>> = _mapScreenCoordinates
+
   private val _screenCoordinates = MutableStateFlow<List<ScreenCoordinate>>(listOf())
   val screenCoordinates: StateFlow<List<ScreenCoordinate>> = _screenCoordinates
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
```diff
@@ -205,15 +210,13 @@ class MapViewModel : ViewModel() {
   }
 
   /**
-   * Get the bottom left and top right corners of the screen in latitude and longitude coordinates.
-   * The corners are calculated based on the center of the screen and the viewport dimensions. If
-   * useBuffer is true, the corners are calculated with a buffer of 2x the viewport dimensions. This
-   * is useful for loading parkings that are not yet visible on the screen.
+   * Updates the bottom left and top right corners of the screen in latitude and longitude
```
LouisJamesVasseur, line 213: Like the new documentation! Much more concise :)

LouisJamesVasseur (approved):
Great fix! Robustly implemented and clear, and the feature works as expected

app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
```diff
@@ -52,6 +52,11 @@ class MapViewModel : ViewModel() {
           } ?: false
         }
 
+  private val _mapScreenCoordinates =
+      MutableStateFlow<Pair<Point, Point>>(
+          Pair(Point.fromLngLat(0.0, 0.0), Point.fromLngLat(0.0, 0.0)))
```
Pijukai, line 57: By experience i can tell you that it is better to use a determined location based on a constant value rather than the (0,0) point which has the tendency of creating unexpected behavior in certain case.

app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
Pijukai: I agree with Louis on this one, documentation is a lot clearer !

app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
matttsss: Yes, I saw at first that no parkings were fetched because of this, I managed to find a fix around this since I found it a bit arbitrary. I nevertheless added a default point in case.

app/src/main/java/com/github/se/cyrcle/ui/map/MapScreen.kt
```diff
@@ -211,8 +219,14 @@ fun MapScreen(
   // Show suggestions screen
   val showSuggestions = remember { mutableStateOf(false) }
 
-  val bitmap = BitmapFactory.decodeResource(LocalContext.current.resources, R.drawable.dot)
-  val resizedBitmap = Bitmap.createScaledBitmap(bitmap, 80, 80, false)
+  val resizedBitmap = remember {
+    val bitmap = BitmapFactory.decodeResource(context.resources, R.drawable.dot)
+    Bitmap.createScaledBitmap(bitmap, 80, 80, false)
```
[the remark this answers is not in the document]
matttsss: Right it wasn't there before but why not xD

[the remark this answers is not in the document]
Pijukai: Thanks !

app/src/main/java/com/github/se/cyrcle/model/map/MapViewModel.kt
Pijukai: Congrats then ! No need to change the point if you managed to handle this.

PR #426, opened 2024-12-16
Update/enhance the e2e tests

Pijukai:
## Content of the PR 

This PR aim to improve and enhance the end to end tests.

## How ?

I implemented the following improvements : 

- For the test `testReviewCard()` which tested only with an anonymous user, I added the part of the test with an Authenticated User to make the test more complete.
- I also completed the test `testProfile()` by adding an assert of the `UserProfileScreen` for an offline user for consistency but also to make the test more relevant. 
- There is a type of helper function in our e2e test file which assert a screen for example `assertUserProfileScreen()`. For those tests, I added asserts on a lot of new component making those functions more up to date.

Finally, I also added a test tag on the add photo button of the `ParkingDetailsScreen` to allow our tests access to it and hence improving test quality.

### DIFF
--- a/app/src/androidTest/java/com/github/se/cyrcle/ui/testEnd2End/MainActivityTest.kt
+++ b/app/src/androidTest/java/com/github/se/cyrcle/ui/testEnd2End/MainActivityTest.kt
@@ -94,6 +94,30 @@ class MainActivityTest {
 
   @Test
   fun testReviewCard() {
+
+    composeTestRule.activity.userRepository.addUser(TestInstancesUser.user1, {}, {})
+
+    // Authenticated user
+
+    authRobot.assertAuthScreen()
+    authRobot.performSignIn()
+
+    mapRobot.assertMapScreen()
+    mapRobot.toList()
+
+    listRobot.assertListScreen(true)
+    listRobot.toCard(0)
+
+    cardRobot.assertParkingDetailsScreen(true)
+    cardRobot.goBack()
+
+    listRobot.toUserProfile()
+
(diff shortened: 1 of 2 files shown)


LouisJamesVasseur (commented):
Before anything else, I'm a little puzzled by the Codacy report. The coverage decreased by 0.57%... any idea why?

Pijukai:
Well to answer your question, I added lines of code in tests but didn't remove any. So it seems weird to me. I think that, since i added the test tag, the formatting added 2 new lines which can potentially explain the decrease. However those lines are tested in my tests so unfortunately I can't tell you why. Seems like a kind of bug on codacy side ?

Pijukai:
Maybe codacy count the tests line as coverable lines and in this case I can't do anything about it.

LouisJamesVasseur:
Thanks for the clarifications. From the diff report on Codacy it seems that you're right. Looks like test tags were added in the process of adding your feature in which case the decreased coverage is understandable

Pijukai:
Well it is still not understandable since I test those lines. If that wasn't the intention then i wouldn't have added the test tag in the first place.

LouisJamesVasseur (approved):
The additions you're providing make sense to me. I like that you added the booleans because it makes sense for some tests to check that a screen is NOT present as well. Overall, this is a robust addendum to the end 2 end tests!

Armetiis (approved):
Thank you for adding relevance to the tests. Although codacy seems to disagree, I like how we are covering more ground

PR #435, opened 2024-12-18
Add Image for each rack type in help screen

Armetiis:
- This PR replaces the placeholder image by an actual picture of each rack type. Note that the images are very compressed but given the size they are displayed, it doesn't make a difference
- The PR also introduces a "Info" icon next to rack type in the Details Screen that navigates to the help screen


matttsss (approved):
All good with me, clear an functional :ok_hand:

LouisJamesVasseur (approved):
Good idea well executed!

Issue #436, opened 2024-12-18
Preview Card shows 1 review instead of 0


Armetiis:
When a parking has 0 reviews, the Preview Card says 1 review

PR #437, opened 2024-12-18
Fix: Number of reviews on PreviewCard 

Armetiis:
- Address #436 

This PR addresses the above mentionned issue by displaying a text "No reviews yet" when there are 0 reviews, and only display the stars when there is 1 review or more

image


andrepeiry (approved):
Seems nice to me !

Issue #407, opened 2024-12-12
Parking data  is out of sync with database when changed.


sean-prz:
There is some changes that are updating the database but not the local cache, making it stale.
For example  :

- Add/Delete a review 
- Add/Delete an Image

If you perform any of this action, it will shows on the UI at first but when navigating between different screens it will disappear. 

Example :
_You can see the review disappearing_
[Screen_recording_20241212_211326.webm](https://github.com/user-attachments/assets/a292d0ef-6025-4684-a43d-70f169b58dc1)

sean-prz:
**This happens because the cache is not updated**

I updated the cache in the following function of the parkingviewmodel : 
`handlenewreview`, `handlereviewdeletion`, `handleReviewUpdate`, and `uploadPicture`

This seems to solve the following problem : 
- adding a review (does not update the score on refresh of the screen) 
- deleting a review (does not update the score on refresh of the screen)
- modifying a review, (does not update the score on refresh of the screen)
- Adding a picture (does not show the picture on refresh of the screen)
@LouisJamesVasseur  There is probably some other case, that I am missing, can you list them if you have any

LouisJamesVasseur:
> **This happens because the cache is not updated**
> 
> I updated the cache in the following function of the parkingviewmodel : `handlenewreview`, `handlereviewdeletion`, `handleReviewUpdate`, and `uploadPicture`
> 
> This seems to solve the following problem :
> 
> * adding a review (does not update the score on refresh of the screen)
> * deleting a review (does not update the score on refresh of the screen)
> * modifying a review, (does not update the score on refresh of the screen)
> * Adding a picture (does not show the picture on refresh of the screen)
>   @LouisJamesVasseur  There is probably some other case, that I am missing, can you list them if you have any

Main issue I've IDed is linked to deletion: deleting a Parking image doesn't reflect in the UI until reload, and deleting a Parking (from admin screen) same thing

sean-prz:
@LouisJamesVasseur AH ok I was not aware of the deletion of a parking issue, I will try to handle that too

LouisJamesVasseur:
> @LouisJamesVasseur AH ok I was not aware of the deletion of a parking issue, I will try to handle that too

It's the issue from my PR last week if you remember

sean-prz:
I am gonna make a quick draft so you can test and see what cases am I missing @LouisJamesVasseur

LouisJamesVasseur:
Sounds good

PR #438, opened 2024-12-18
fix: stale cache, displays out of sync data

sean-prz:
- fix : #407 

#### maintains  the displayed informations  in sync with the datas.
In particular : 
- Adding/editing/Deleting a review is now reflected persistently in the ui
- Adding/removing a picture is reflected persistently in the ui
- Deleting a parking is reflected persistently in the ui

## How
Add a private function `updateCache` that does that. 
call this function every time the viewmodel makes a change that is passed to a repo.

## Why
To keep the UI in sync with data


LouisJamesVasseur (approved):
I've tested most of the problematic edge cases I ran into and your fix fixes all of these. Great work!

Armetiis (approved):
Good work. Thanks

Issue #365, opened 2024-12-06
Authentication Failure on new phones


matttsss:
Upon connection from a phone with no credentials, authentication fails since the authenticator does not find any credentials.

PR #439, opened 2024-12-19
Attemp to bugfix missing credentials

matttsss:
## What is this PR?
This PR aims at bug-fixing an issue where the prompt to enter your credentials does not show up when credentials are missing. There is a trace of this issue being present in this github thread: https://github.com/android/identity-samples/issues/53

The issue in itself is not solved in the app, but this is the best try to apply the Google Credential Manager tutorial (first ask with `.setFilterByAuthorizedAccounts(true)` and then `.setFilterByAuthorizedAccounts(false)` if there were no credentials).

## How?
For this, the code for making the credential request was moved into a "semi-recursive" helper function. The request is then made once with `.setFilterByAuthorizedAccounts(true)` and then recurses calling this time with `setFilterByAuthorizedAccounts(false)` then if authentication still fails, it calls the `onFailure` with the correct exception.

Concerns #365

### DIFF
--- a/app/src/main/java/com/github/se/cyrcle/model/authentication/AuthenticationRepositoryGoogle.kt
+++ b/app/src/main/java/com/github/se/cyrcle/model/authentication/AuthenticationRepositoryGoogle.kt
@@ -50,6 +50,14 @@ constructor(
     makeRequest(onSuccess, onFailure, true)
   }
 
+  /**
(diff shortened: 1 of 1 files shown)


Pijukai (approved):
Thank you for trying until the last moment to fix the bug. Using the last version of the credential manager of Google is a good idea and your fix is really creative so great Job! 
Approved !

